PR #1, opened 2023-05-04
Versioning poc

airportyh:
# Why?

We need to implement a versioning scheme that'll work in the short term and in the long term.
See https://replit.com/@util/Design-docs#goval/nixmodules_versions.md for more context.

## Changes

1. docs
2. added `lock_modules.py` script for generating the module registry and pinning module versions
3. added `id` and `community-version` fields to module definition
4. moved module definition list to `pkgs/modules/default.nix`
5. moved modules to subdir `pkgs/modules`
6. added pruneVersion utility to get first 2 numbers of a semver, used to get the community version for modules
7. added prybar. Depends on https://github.com/replit/prybar/pull/108


README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+`modules.json` is similar to a lock file in used in common packagers in that it fixes
+the exact version of each module. This file looks something like:
+
+```json
+{
+  "modules": {
+    "nodejs-18.12-m1.1": {
+      "commit": "4ec006c0eb247320e77c0abbf46b6f9e33370f81",
```
ryantm, line 66: We are going to have to disallow rebase merging. Otherwise, the commits we save in the the lockfile might not end up in the registry. Another solution would be to generate the registry automatically in CI on the main branch.

flake.nix
```diff
@@ -2,20 +2,32 @@
...
       overlays.default = final: prev: {
         moduleit = self.packages.${prev.system}.moduleit;
       };
       formatter.x86_64-linux = pkgs.nixpkgs-fmt;
-      packages.x86_64-linux = import ./pkgs { inherit pkgs self nixpkgs nixmodules-stable; };
+      packages.x86_64-linux = import ./pkgs {
+        inherit pkgs self nixpkgs nixmodules-stable; 
+      };
+      devShells.x86_64-linux.default = pkgs.mkShell {
+        buildInputs = [
```
ryantm, line 25:
```suggestion
        packages = [
```

I think that's the preferred name here these days.

pkgs/default.nix
```diff
@@ -3,31 +3,19 @@
 with pkgs.lib;
 
 let
-  mkModule = path: pkgs.callPackage ./moduleit/entrypoint.nix {
-    configPath = path;
+  modules = import ./modules {
```
ryantm, line 6: this is importing modules in a second place in addition to the top level outputs.modules. I'd prefer to pass `self` into the pkgs evaluation and get the modules from `self.modules`

pkgs/moduleit/entrypoint.nix
```diff
@@ -1,7 +1,14 @@
 { pkgs ? import <nixpkgs> { }
 , configPath
 }:
-
+let pruneVersion = version:
+  let
+    parts = pkgs.lib.strings.splitString "." version;
+    major = builtins.elemAt parts 0;
+    minor = builtins.elemAt parts 1;
```
ryantm, line 8: https://github.com/NixOS/nixpkgs/blob/master/lib/versions.nix there are already major and minor functions you could use here.

`pruneVersion = lib.majorMinor`

pkgs/modules/go/default.nix
```diff
@@ -1,6 +1,17 @@
 { pkgs, ... }:
+let
+  goversion =
```
ryantm, line 3:
```suggestion
  goversion = lib.majorMinor pkgs.go.version
```

ryantm (approved):

pkgs/moduleit/entrypoint.nix
airportyh: :+1: I thought there might be.

README.md
ryantm: It changes the git commit SHAs when it rewrites the history.

PR #2, opened 2023-05-04
Mass migration

airportyh:
# Why?

Move all Nix modules from nixpkgs-replit so they'll be easier to maintain here. Also, we want to do this sooner than later because the `lock_modules.py` scheme make references to git commit shas.

## Changes

moved every module...


pkgs/modules/bun/bun.nix
cdmistman, line 1: to keep dir structure consistency, this should probably be `pkgs/bun/default.nix`

pkgs/modules/bun/default.nix
```diff
@@ -0,0 +1,45 @@
+{ pkgs, pruneVersion, ... }:
+
+let
+  bun = pkgs.callPackage ./bun.nix { };
```
ryantm, line 4: Yes, and this should come via an overlay we make instead of callPackage here.

ryantm: I can do that refactoring in a followup pass though.

ryantm (approved):

PR #4, opened 2023-05-05
Bundle registry

airportyh:
# Why?

Automatically create a disk that contains all packages needed by Nix modules

## Changes

1. added python script and associated shell script to do the bundling
2. the script outputs to the `./build` directory which is added to .gitignore

## Testing

```
nix develop
python lock_modules.py
python bundle_registry.py
```


pkgs/modules/nodejs/default.nix
```diff
@@ -19,9 +19,9 @@ in
 
 {
   id = "nodejs";
-  name = "Node.js Tools";
+  name = "Node.js Tools BLAH";
```
ryantm, line 22: some debugging stuff?

ryantm (approved):

PR #8, opened 2023-05-09
bundle: add bundle-locked output generated from lockfile

ryantm:
Why
===
* Being able to do the building in Nix makes it easier to integrate with other tools.

What changed
===
* Added a new output bundle-locked, which is the bundle generated from the lockfile


airportyh (approved):
beautiful!

PR #9, opened 2023-05-09
bundle-image: remove stable input; add lockfile to bundle-image

ryantm:
Why
===
* If we have lockfiles and aliases, it doesn't seem necessary to also have Stable and Beta registries.
* We want to get the new lockfile in the image

What changed
===
* Removed nixmodules-stable input and everything depending on it
* Add modules.json to bundle-image at path /etc/nixmodules/modules.json

Test plan
===
* `nix build .#bundle-image` worked
* will check to make sure the lockfile is available

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -15,25 +15,6 @@
         "type": "github"
       }
     },
-    "nixmodules-stable": {
-      "inputs": {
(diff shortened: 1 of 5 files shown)


airportyh (approved):
wfm

PR #13, opened 2023-05-11
Upgrade maps

airportyh:
# Why?

Move to new map-based upgrade scheme: https://replit.com/@util/Design-docs#goval/nixmodules_versions.md

## Changes

1. Added upgrade maps definition
2. Added check_upgrade_maps.py script to check for cycles and invalid module ID references
3. added `auto-upgrade.json` and `recommend-upgrade.json` plus their shas into the bundle image
4. added `active-modules.json` to the bundled images for use in store listings
5. removed aliases section from module registry
6. regenerated the registry
7. added name, and description to the output of `nix eval .#modules --json`
8. changed the id in the output json as the fully specified module ID


pkgs/upgrade-maps/default.nix
```diff
@@ -0,0 +1,15 @@
...
+    "nodejs-19.1-m1.1" = "nodejs-19.1-m1.2";
+  };
+
+  recommend = {
+    "nodejs-14.21-m1.2" = "nodejs-14.21-m2.0";
+    "nodejs-16.18-m1.2" = "nodejs-16.18-m2.0";
+    "nodejs-18.12-m1.2" = "nodejs-18.12-m2.0";
+    "nodejs-19.1-m1.2" = "nodejs-19.1-m2.0";
+  };
+}
```
ryantm, line 15: If we wrote this like:

```nix
{
    "nodejs-14.21-m1.1" = {to = "nodejs-14.21-m1.2"; auto = true;};
    "nodejs-16.18-m1.1" = {to = "nodejs-16.18-m1.2";  auto = true;};
    "nodejs-18.12-m1.1" = {to = "nodejs-18.12-m1.2"; auto = true;};
    "nodejs-19.1-m1.1" = {to = "nodejs-19.1-m1.2"; auto = true;};
    "nodejs-14.21-m1.2" = {to = "nodejs-14.21-m2.0"; };
    "nodejs-16.18-m1.2" = {to = "nodejs-16.18-m2.0"; };
    "nodejs-18.12-m1.2" = {to = "nodejs-18.12-m2.0"; };
    "nodejs-19.1-m1.2" = {to = "nodejs-19.1-m2.0"; };
  };
}
```

Then nix would enforce the requirement that each module can only have one place it goes to.

pkgs/moduleit/module-definition.nix
```diff
@@ -580,7 +580,7 @@ in
       let
 
         moduleJSON = {
-          id = config.id;
+          id = "${config.id}-v${config.version}";
```
ryantm, line 583: This is going to be confusing because the id people specify doesn't match the id used in the json.

There is something similar to this in nixpkgs. They have a derivation "name", which is defined as `"${pname}-${version}"`.

We either should use a different name for one of these, or figure out how to get rid of one of them.

flake.nix
```diff
@@ -35,5 +35,8 @@
       modules = import ./pkgs/modules {
         inherit pkgs;
       };
+      upgrade-maps = import ./pkgs/upgrade-maps {
+        inherit pkgs;
+      };
```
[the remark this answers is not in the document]
airportyh: It's being used in `check_upgrade_maps.py` for verification. Is that a valid use case? Seem like having to build a file is a bit heavy for that.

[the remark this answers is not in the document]
ryantm: You can make a new "checks" flake output that runs your python check program and pass in the upgrade map that way. This will also make it run when you do `nix flake check`.

pkgs/moduleit/module-definition.nix
airportyh: @masad-frost favors removing version altogether (I guess move to the version is implicitly inside the ID). But I think that'd be weird for author to have to change their package's ID each time before they publish an update. Hmm, not sure...

airportyh: Maybe call it `registryId`; or just revert this change - you can always concatenate the id + version to get the same thing.

flake.nix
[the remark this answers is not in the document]
airportyh: :+1:

pkgs/moduleit/module-definition.nix
masad-frost: I have a strong opinion that we shouldn't have 2 things we refer to as version and one of them has no real versioning concept/is not part of the versioning system.

airportyh: @masad-frost I think you are saying the upgrade maps system removes the need for the version field?

Although the upgrade maps implements a versioning system, I don't think of it as overlapping with the term "version", because I associate the word with numeric versions or some sort of tag. Even when using the maps, we'd still need to use a versioning scheme in addition to it, whether numeric, semver-based, date-based, or sha-based. It's true that we could allow each author to independently choose their own versioning scheme.

I'd like to think through some alternatives from the authoring perspective. But first, I'll summarize the current design.

## Current Design

The IDs within the registry (let's call this the *registry module ID*) look like:

`nodejs-18-v1.2` which is composed of `<module ID>-v<version>`

In this case the "community version" is embedded in the module ID, but that's up to the module author, and it's no longer part of our system. The author writes something like:

```
{
  id = "nodejs-18";
  name = "Node.js 18 Tools";
  version = "1.2";
  ...
}
```

We have a lock_modules.py script that pins each unique module registry ID to a Nix output path. If a change is detected, the author is forced to increment the version number before publishing. (Is this script necessary? More below...)

## Alternative 1

Having both id and name is probably redundant and we can probably have just name. So the
author would write:

```
{
  name = "nodejs-18";
  version = "1.2";
  ...
}
```

An ID field can be generated using `"${name}-v${version}"`, and we can use the "description" field for more wordy user-facing description of the module.

*Pro:* familiar to users of popular packages like npm

## Alternative 2

If we remove the version field altogether, it puts the onus on the user to choose a versioning scheme:

```
{
  id = "nodejs-18-1.2";
}
```

or


```
{
  id = "nodejs-18-20230518";
}
```

or

```
{
  id = "nodejs-18-8ca571d3efa03de178c4262c5b908ef1e3ff2218";
}
```

Substitute "name" for "id" if you prefer.

Again, a lock_modules.py script can be used to pin each registry ID to an output path. If something changed, the user would be forced to change the ID.

We may want to expose the upgrade map system to users at some point, so it'd serve them well to use a scheme that's easy to work with when adding entries to and editing the map.

*Pro:* more user choice; less structure in our system
*Con:* less uniformity among modules

## Alterative 3

We could chose to just use git to generate the version tags, in which case the user will not need to concern themselves with updating the version when they change code:

```
{
  id = "nodejs-18";
}
```

The lock_modules.py script would generate registry IDs like: `nodejs-18-<commit sha>`. Or it can be date-based: `nodejs-18-202305180934`, based on the timestamp of the commit.

*Pro:* no need to maintain version field or id field
*Con:* the commit sha-based version scheme would be hard to work with when editing the upgrade maps.
Date-based would be better, but not as easy as a numeric scheme.

## Do we need the lock_modules.py script?

This script gives these things:

1. listing the build inputs of all historical versions which is needed by the current Nix modules disk scheme to build them and put them on the disk
2. the same listing is used in the module/extensions store UI
3. ensures the same registry module ID cannot generate multiple different Nix output paths in different versions of the repo

It comes with these problems:
1. developer maintenance overhead (update / merging)
2. it's an additional layer. Do we push it onto community authors as well?

we can maybe get away with not using this script:

* 1 and 2 can be done by go through every historical commit in the repo and evaluating (not building) every module derivation to
get all possible `<module ID>-<commit ID>` -> Nix output path mappings. This could be expensive if the repo has many commits and/or is large in size.
* 3 becames unnecessary
* alternative 3 lends itself better to this approach because conflicting entries is not possible

## Community Modules

We haven't talked about how to handle updates (auto and recommended) with community Nix modules, but maybe that should factor into the design here
* should they be able to specify upgrade maps?
* how to list their modules in the store?
* to they open a PR in our repo or publish their own repo?
* can users just put the git URL of a 3rd party repo + build target in the `modules` array in `.replit`?

## Other Alternatives?

If you think of other alternatives, please share.

## Analysis

Additional pros/cons of the current design vs each alternative?

airportyh: One more thing to throw it for consideration: using upgrade maps for auto-upgrading minor and patch versions could get tedious if there are frequent updates to a module, and we'd also end up with many entries in the map. We can consider using upgrade maps only for recommended upgrade, but a semver-based for auto upgrades. Or coming up with automation to add entries to the map for the auto upgrade case.

airportyh: Had a discussion and agreement last Friday, but to have this in written form:

1. We decided on going with alternative 3
2. And we are keeping lock_modules.py, which will take on the job of automatically generating module registry IDs in the form of `v<version>-<date>-<commit>` where:
    * `version` is a numeric auto-incremented number that starts at 1
    * `date` is an 8-digit sequence of the form: `YYYYMMDD`
    * `commit` is a 7-digit prefix of the git commit sha

More details will be in the docs.

pkgs/moduleit/module-definition.nix
```diff
@@ -481,23 +481,13 @@ in
...
-    };
-
     packages = mkOption {
       type = types.listOf types.package;
       default = [ ];
       description = "The set of packages to appear in the repl.";
     };
 
     replit = {
-      version = mkOption {
```
airportyh, line 496: This was unused and I thought we might as well remove it too.

README.md
```diff
@@ -13,125 +13,107 @@ nix eval .#modules --json | jq
...
+ "bun-0.5:v1-20230522-ec43fbd": {
+   "commit": "ec43fbd5f1ad8556bb64da7f77ae4af8d9ae6461",
+#    "path": "/nix/store/l08f5vl1af9rpzd7kvr0l5gx9v7y8p12-replit-module-bun-0.5"
+#  },
+#  "c-clang14.0:v1-20230522-ec43fbd": {
+#    "commit": "ec43fbd5f1ad8556bb64da7f77ae4af8d9ae6461",
+#    "path": "/nix/store/n4vzd9rkpjs72xj9yvlakxh3bardvdki-replit-module-c-clang14.0"
+#  },
+#  ...
+# }
```
ryantm, line 78:
```suggestion
```

README.md
```diff
@@ -13,125 +13,107 @@ nix eval .#modules --json | jq
...
+nix build .#bundle-squashfs
 ```
 
-To move forward, you'll have to increment the version of the associated module.
+## Lock Modules
 
-See more about the versioning scheme at: https://replit.com/@util/Design-docs#goval/nixmodules_versions.md
+`scripts/lock_modules.py` generates/updates the module registry file `modules.json`.
+modules.json is similar to a lock file used in common packagers in that it fixes
+the exact version of each module. It looks like:
```
ryantm, line 66:
```suggestion
```

pkgs/active-modules/default.nix
```diff
@@ -0,0 +1,85 @@
...
+    all-modules-list;
+in
+pkgs.stdenv.mkDerivation {
+  pname = "active-modules";
+  version = "1";
+  src = ./.;
+  installPhase = ''
+  echo '${builtins.toJSON active-modules-registry}' > $out
+  '';
+  meta = {
```
ryantm, line 80:
```suggestion
  passthru = {
```

meta is supposed to be for metadata about the package. passthru should do what you want.

pkgs/active-modules/default.nix
```diff
@@ -0,0 +1,85 @@
...
+            };
+          }
+    )
+    {}
+    all-modules-list;
+in
+pkgs.stdenv.mkDerivation {
+  pname = "active-modules";
+  version = "1";
+  src = ./.;
```
ryantm, line 76:
```suggestion
  dontUnpack = true;
  dontConfigure = true;
  dontBuild = true;
```

pkgs/upgrade-maps/default.nix
```diff
@@ -0,0 +1,37 @@
...
+}) entries;
+filter-auto = filterAttrs (mod: entry: entry.auto or false);
+filter-recommend = filterAttrs (mod: entry: !(entry.auto or false));
+auto = present-entries (filter-auto mapping);
+recommend = present-entries (filter-recommend mapping);
+in
+pkgs.stdenv.mkDerivation {
+  pname = "upgrade-maps";
+  version = "1.0.0";
+  src = ./.;
```
ryantm, line 24:
```suggestion
  dontUnpack = true;
  dontConfigure = true;
  dontBuild = true;
```

pkgs/upgrade-maps/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  installPhase = ''
+  mkdir $out
+  echo '${builtins.toJSON auto}' > $out/auto-upgrade.json
+  echo '${builtins.toJSON recommend}' > $out/recommend-upgrade.json
+  '';
+  meta = {
```
ryantm, line 30:
```suggestion
  passthru = {
```

scripts/check_upgrade_maps.py
```diff
@@ -0,0 +1,61 @@
...
+# It should be run in CI to ensure we don't merge in mistakes.
+
+import os
+import json
+import subprocess
+
+module_registry_file = 'modules.json'
+
+def get_upgrade_maps():
+  output = subprocess.check_output(['nix', 'eval', '.#upgrade-maps.meta.auto', '--json'])
```
ryantm, line 19:
```suggestion
  output = subprocess.check_output(['nix', 'eval', '.#upgrade-maps.auto', '--json'])
```

scripts/check_upgrade_maps.py
```diff
@@ -0,0 +1,61 @@
...
+  auto = json.loads(output)
+  output = subprocess.check_output(['nix', 'eval', '.#upgrade-maps.meta.recommend', '--json'])
```
ryantm, line 21:
```suggestion
  output = subprocess.check_output(['nix', 'eval', '.#upgrade-maps.recommend', '--json'])
```

pkgs/active-modules/default.nix
```diff
@@ -0,0 +1,85 @@
...
+  installPhase = ''
+  echo '${builtins.toJSON active-modules-registry}' > $out
+  '';
+  meta = {
+    # this allows you to query for this info wo building it:
+    # nix eval .#active-modules.meta.info --json
```
ryantm, line 82:
```suggestion
    # nix eval .#active-modules.info --json
```

ryantm (approved):

PR #14, opened 2023-05-12
Removed community versions

airportyh:
## Why?

From our discussions we decided that community versions should not be a formal thing baked into the module system.

## Changes

1. Removed community-version field in module definitions
2. update all modules versioning scheme. Most of them embed the "community version" into their ID. some do not have one.
3. Instead of `m` prefix to a module version, I switched it to `v`


ryantm (approved):
I think we should probably put the community version in all the names, not just the ones whose version we parameterize.

PR #15, opened 2023-05-15
[bundle] ensure bundle-image and bundle-image-tarball have no dependencies

ryantm:
Why
===
* Nix was scanning the bundle image/tarball and finding references which you could see by running

nix path-info -rsSh .#bundle-image

these references ended up in the runtime closure, which means they could not be garbage collected without also collecting the image itself.

What changed
===
* Use the experimental feature unsafeDiscardReferences enabled under experimental flag discard-references to tell Nix that really these derivations has no references in the end.

Test plan
===
* Confirmed nix build works for both image and tarball
* Confirmed they do not have references

### DIFF
--- a/pkgs/bundle-image-tarball/default.nix
+++ b/pkgs/bundle-image-tarball/default.nix
@@ -1,11 +1,18 @@
-{runCommand, lib, revstring, bundle-image, pigz }:
+{system, bash, lib, bundle-image, revstring, coreutils, gnutar, pigz} :
 
-let
-  label = "nixmodules-${revstring}";
-in
-
-runCommand label {} ''
-  echo "making tarball..."
-  mkdir -p $out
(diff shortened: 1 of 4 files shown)


pkgs/bundle-image-tarball/default.nix
```diff
@@ -1,11 +1,18 @@
...
-let
-  label = "nixmodules-${revstring}";
-in
-
-runCommand label {} ''
-  echo "making tarball..."
-  mkdir -p $out
-  tar --use-compress-program=${pigz}/bin/pigz -Scf $out/disk.raw.tar.gz -C ${bundle-image} disk.raw
-''
+derivation {
```
airportyh, line 3: Is derivation a "statement"? Haven't seen it before.

airportyh (approved):
Nice! Tested to work.

pkgs/bundle-image-tarball/default.nix
ryantm: It's a builtin function https://nixos.org/manual/nix/stable/language/derivations.html

PR #16, opened 2023-05-16
bundle-squashfs: add output

ryantm:
Why
===
* For local and CI, it is useful to have a small squashfs drive

What changed
==
* Added bundle-squashfs output

Test plan
===
* nix build .#bundle-squashfs works
* when mounting it, it looks like it has the correct files


airportyh (approved):
Nice!

PR #18, opened 2023-05-19
Updated rust to 1.69; added nixpkgs-unstable

airportyh:
# Why?

Upgrade rust to get their new registry updating scheme that will be fast.

# Changes

1. added nixpkgs-unstable as input
2. rust uses things from the unstable channel

# Testing

* `nix build .#bundle`
* `ls result` and see rust at the new version

Disk usage didn't seem to go up much:

```
nix build .#bundle-image
$ ls -l result/disk.raw 
-r--r--r-- 1 toby toby 22548578304 Dec 31  1969 result/disk.raw
```

### DIFF
--- a/pkgs/moduleit/entrypoint.nix
+++ b/pkgs/moduleit/entrypoint.nix
@@ -1,6 +1,6 @@
 { pkgs ? import <nixpkgs> { }
+, pkgs-unstable ? import <nixpkgs-unstable> { }
 , configPath
-, pkgs-unstable
 }:
 (pkgs.lib.evalModules {
   modules = [
--- a/pkgs/moduleit/entrypoint.nix
+++ b/pkgs/moduleit/entrypoint.nix
@@ -1,5 +1,6 @@
 { pkgs ? import <nixpkgs> { }
 , configPath
+, pkgs-unstable
 }:
 (pkgs.lib.evalModules {
   modules = [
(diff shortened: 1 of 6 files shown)


pkgs/moduleit/entrypoint.nix
```diff
@@ -1,5 +1,6 @@
 { pkgs ? import <nixpkgs> { }
 , configPath
+, pkgs-unstable
```
ryantm, line 3: Is this going to be problematic that source-based module compiles need to provide this argument?

airportyh: if they write { pkgs, ... } then the code will still work. I think it's ok.

ryantm: This line will break: https://github.com/replit/nixmodules/blob/9dc9ea9f869597e2a1363a197445165d0aaded61/pkgs/moduleit/moduleit.sh#L13 because it doesn't provide pkgs-unstable. It is okay that it doesn't provide pkgs, because we've provided a default for that.

Potential fix: change the rust module to just use pkgs but override the definition of pkgs for the purposes of the rust module. For example, maybe:
```nix
{
  mkUnstableModule = path: pkgs-unstable.callPackage ../moduleit/entrypoint.nix {
    configPath = path;
  };
}
```

ryantm: Other possible fix: provide a default of pkgs-unstable.

airportyh: I put `import <nixpkgs-unstable> { }` as default for pkgs-unstable. I think this should work as long as you have a channel configured named `nixpkgs-unstable` which they Repls do. Does this work?

ryantm (approved):

PR #19, opened 2023-05-22
C major community version

airportyh:
# Why?

We should be tracking Clang on only it's major version.

https://releases.llvm.org/


modules.json
```diff
@@ -1,90 +1,90 @@
 {
```
cdmistman, line 1: Should this file be marked binary? It's just generated output right?

pkgs/modules/c/default.nix
```diff
@@ -10,7 +10,7 @@ let
   dap-cpp = pkgs.callPackage ../../dap-cpp { };
   dap-cpp-messages = import ../../dap-cpp/messages.nix;
 
-  clang-version = lib.versions.majorMinor clang.version;
+  clang-version = lib.versions.major clang.version;
```
cdmistman, line 13: Thought: should we have different modules for the different C standards? I'd ship C99 in its own module vs C11 vs...

cdmistman (approved):

pkgs/modules/c/default.nix
airportyh: I think that's a flag to clang. Maybe should be a configuration? But we don't have a way for users to configure things currently...

modules.json
airportyh: It can be mergeable by humans.

pkgs/modules/c/default.nix
cdmistman: gotcha, we should look into adding something for this...

PR #20, opened 2023-05-24
Add aliases

airportyh:
# Why?

Add these aliases to make existing beta templates work.

## Changes

1. added alias
2. ran lock command which was forgotten after https://github.com/replit/nixmodules/pull/18

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -3,77 +3,153 @@
     "commit": "94a8f12da8d04565c302689dadcdddc1f812386f",
     "path": "/nix/store/asz1nzgcgnxh9nl9b1f9k08v6zwf9j6x-replit-module-bun-0.5"
   },
+  "bun-0.5:v2-20230523-b8111f7": {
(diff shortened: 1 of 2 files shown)


modules.json
cdmistman, line 1: considering this file will likely be updated after every PR, i think it'd be a good idea to mark this as binary and ban manual edits to the file

i'm not sure the best way to accomplish this now, but i suspect something like this would work:
- ensure the file is well-formatted
- ensure the line count diff is as-expected (4 per module)
- run the lock again and ensure no contents change

i'm also not convinced that the commits here will be correct? commits to `main` will be squashed so the commit in the object will never exist on `main` - perhaps we should only serve/change this file in CD?

cdmistman (approved):

modules.json
airportyh: We've turned off squash commit and rebasing for this repo to prevent this. My initial thought was to let people hand merge it. But I can see point against that.

PR #22, opened 2023-05-24
Fix upgrade maps; Make changelog optional

airportyh:
# Why

build upgrade-maps failed

# What changed

Made changelog field optional


ryantm (approved):
I think it is confusing to write attrsets that look like another languages if statement bodies, but otherwise looks fine.

PR #26, opened 2023-05-25
lock_modules.py verifies commits exist

airportyh:
# Why?

As I have shown, it's possible to accidentally squash the commit(s) used to generate the modules.json file

## Changes

Added code in lock_modules.py to verify all commits still exist, with the `-v` flag.


ryantm (approved):
Yay thanks

PR #32, opened 2023-05-30
More CI checks

airportyh:
# Why?

Want CI check to actually test build the newly updated modules, but not everything because I am worried that'll take it over our disk usage limit.
Add a check to guard against hand-merging of modules.json where multiple entries have the same numeric version.

## Changes

1. moved commit verification logic out of `lock_modules.py` into `check_modules.py` to reduce responsibility of the former
2. added numeric version duplicate check to `check_modules.py`
3. added `build_changed_modules.py` script to build modules/versions that were added vs main
4. call the new checks in `ci_check.sh`

### DIFF
--- a/scripts/build_changed_modules.py
+++ b/scripts/build_changed_modules.py
@@ -1,0 +1,54 @@
+# This script test builds the updated modules for a pull request branch.
+
+import json
+import os
+import subprocess
+import argparse
+
+module_registry_file = 'modules.json'
+
+def get_modules():
+  if not os.path.isfile(module_registry_file):
+    return {}
+  f = open(module_registry_file, 'r')
+  modules = json.load(f)
+  f.close()
+  return modules
+
+def build_module(module_id):
+  args = ['nix', 'build', '.#modules."%s"' % module_id]
+  print(" ".join(args))
+  subprocess.run(args)
+
+def get_upstream_modules(branch):
+  args = ['git', 'show', '%s:modules.json' % branch]
+  print(" ".join(args))
+  output = subprocess.check_output(args)
+  return json.loads(str(output, 'UTF-8'))
+
+def main():
+  parser = argparse.ArgumentParser(
+    prog='build_changed_modules',
+    description='builds the modules that were changed from an upstream branch',
+  )
+
+  parser.add_argument('upstream_branch')
+
+  args = parser.parse_args()
+
+  upstream_modules = get_upstream_modules(args.upstream_branch)
+  upstream_ids = set(upstream_modules.keys())
+  current_modules = get_modules()
+  current_ids = set(current_modules.keys())
+  new_modules = current_ids - upstream_ids
+  for module_registry_id in new_modules:
+    module_id, _ = module_registry_id.split(':')
+    build_module(module_id)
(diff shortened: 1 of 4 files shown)


scripts/check_modules.py
```diff
@@ -0,0 +1,59 @@
...
+def check_commits_exist():
+  for commit in get_commits():
+    assert commit_exists(commit), 'Commit %s does not exist!' % commit
+
+def get_modules():
+  if not os.path.isfile(module_registry_file):
+    return {}
+  f = open(module_registry_file, 'r')
+  modules = json.load(f)
+  f.close()
```
cdmistman, line 40: nit: should probably open the file once then use that in both checks. otherwise once `modules.json` gets large (and it will get large) this script will take *forever* and potentially OOM (depending on python GC)

scripts/build_changed_modules.py
```diff
@@ -0,0 +1,56 @@
...
+    module_id, _ = module_registry_id.split(':')
+    build_module(module_id)
+    # verify output is same as entry
+    actual_path = os.readlink('result')
+    referenced_path = current_modules[module_registry_id]['path']
+    assert actual_path == referenced_path, 'output path for %s does not match: %s vs %s' % (module_registry_id, actual_path, referenced_path)
+    print('%s ok' % module_registry_id)
+
+if __name__ == '__main__':
+  main()
```
cdmistman, line 56:
```suggestion
  main()

```

scripts/check_modules.py
```diff
@@ -0,0 +1,59 @@
...
+    if short_key in uniques:
+      raise Exception("%s has duplicate numeric versions: %s and %s" % (module_id, key, uniques[short_key]))
+    uniques[short_key] = key
+
+def main():
+  check_commits_exist()
+  check_no_duplicate_numeric_versions()
+
+if __name__ == '__main__':
+  main()
```
cdmistman, line 59:
```suggestion
  main()

```

scripts/ci_check.sh
```diff
@@ -14,5 +14,8 @@ cat result
 nix develop $NIX_FLAGS
 echo "Verify modules.json"
 python scripts/lock_modules.py -v
+python scripts/check_modules.py
 echo "Verify upgrade maps"
 python scripts/check_upgrade_maps.py
+echo "Build new/updated modules"
+python scripts/build_changed_modules.py main $(git branch --show-current)
```
cdmistman, line 21:
```suggestion
python scripts/build_changed_modules.py main $(git branch --show-current)

```

scripts/build_changed_modules.py
```diff
@@ -0,0 +1,56 @@
...
+    description='builds the modules that were changed from an upstream branch',
+  )
+
+  parser.add_argument('upstream_branch')
+  parser.add_argument('current_branch')
+
+  args = parser.parse_args()
+
+  checkout(args.upstream_branch)
+  upstream_modules = get_modules()
```
cdmistman, line 40: you can `git show <rev> modules.json` - better imo since it doesn't mutate the environment

cdmistman (approved):
besides suggestions, lgtm

PR #34, opened 2023-06-05
Add ts extensions for nodejs LSP

airportyh:
# Why?

1. Node.js module LSP does not support .ts files.
2. Setup environment variables same as https://replit.com/@replit/Nodejs to allow global installs, etc.

## Change

1. Added extensions override for typescript language server.
2. Added env var setup.

### DIFF
--- a/pkgs/modules/nodejs/default.nix
+++ b/pkgs/modules/nodejs/default.nix
@@ -32,6 +32,8 @@ in
 
   replit = {
 
+    languageServers.typescript-language-server.extensions = [".js" ".jsx" ".ts" ".tsx" ".json"];
+
     runners.nodeJS = {
       name = "Node.js";
       language = "javascript";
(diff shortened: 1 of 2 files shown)


pkgs/modules/nodejs/default.nix
```diff
@@ -32,6 +32,8 @@ in
 
   replit = {
 
+    languageServers.typescript-language-server.extensions = [".js" ".jsx" ".ts" ".tsx" ".json"];
```
cdmistman, line 35: i think it'd be better to set this in the typescript language server module, no?

can probably just set `language = "typescript"` instead of `javascript` there too, but idk what extensions that covers

ryantm: I think this is missing an `(import ../typescript-language-server ...`

airportyh: in config.toml, javascript is `[".js", ".es6", ".mjs", ".cjs"]`, typescript is `[".ts", ".mts", ".cts"]`.

https://github.com/replit/repl-it-web/blob/main/app/filetypes/config.toml

The problem is this is a union of the two. Maybe it should be `languages` , not language.

But the problem of multiple instance of ts-language-server when you have multiple modules that use it still stands. Not sure of a clean solution.

ryantm (changes requested):

pkgs/modules/nodejs/default.nix
airportyh: > I think this is missing an `(import ../typescript-language-server ...`

That's on line 24.

ryantm (approved):

PR #37, opened 2023-06-06
formatting: run nix fmt, and add nix fmt check

ryantm:
Why
===
* no style wars

What changed
===
* ran `nix fmt`
* added `nix fmt -- --check .` to the ci_checks scripts

Test plan
===
* Tested ci_checks script detects formatting problems
* Tested ci_checks script completes successfully

### DIFF
--- a/pkgs/active-modules/default.nix
+++ b/pkgs/active-modules/default.nix
@@ -12,63 +12,66 @@
 
 # or
 # nix eval .#active-modules.meta.info --json | jq
-  
+
 { self, pkgs }:
 with pkgs.lib;
 let
   active-modules = self.modules;
   get-version = (tag:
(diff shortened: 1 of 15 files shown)


pkgs/active-modules/default.nix
```diff
@@ -12,63 +12,66 @@
...
-          inherit (module-info) name description;
-        };
-      in
-      if ! builtins.hasAttr module-id active-modules then
-        registry
-      else
-        if builtins.hasAttr module-id registry then
-          let prev-entry = registry.${module-id};
+  active-modules-registry =
+    foldr
```
masad-frost, line 32: excuse me, can you use tabs instead of spaces /s

masad-frost (approved):

PR #39, opened 2023-06-07
Th pylsp wrapper

airportyh:
Why
===

If you use the Python Nix module, and then use poetry to install ujson, it breaks pylsp because it requires ujson as well, and `.pythonlibs` being in `PYTHONPATH` affects it and causes it not to be able to find stdc++.so

**Note: this fix depends on https://github.com/replit/repl-it-web/pull/32128**

What changed
============

1. use a wrapper for pylsp to freeze `PYTHONPATH` for it
2. add lsp configuration for jedi to add `.pythonlibs` to its path. see https://github.com/python-lsp/python-lsp-server/blob/develop/CONFIGURATION.md

Test plan
=========

The following test is to be performed after nixmodules disk goes into Canary and https://github.com/replit/repl-it-web/pull/32128 gets merged.

1. make blank repl
2. use module tool to add the python module
3. `poetry init` and step through the prompt to get project inited
5. `poetry add requests`
4. `poetry add ujson`
6. in `main.py` write
```
import requests
requests.get()
```
7. hover over word "get" and see hover help
8. `poetry add ujson`
9. `kill 1`
10. refresh browser
11. hover over word "get" and still see hover help

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -3,6 +3,8 @@
 let
   community-version = lib.versions.majorMinor python.version;
 
+  pylibs-dir = ".pythonlibs";
+
   pip = pkgs.callPackage ../../pip {
     inherit pypkgs;
   };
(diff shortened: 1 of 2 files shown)


pkgs/modules/python/default.nix
```diff
@@ -161,7 +175,7 @@ in
   };
 
   replit.env =
-    let userbase = "$HOME/$REPL_SLUG/.pythonlibs";
+    let userbase = "$HOME/$REPL_SLUG/${pylibs-dir}";
```
ryantm, line 178:
```suggestion
    let userbase = "$REPL_HOME/${pylibs-dir}";
```

pkgs/modules/python/default.nix
```diff
@@ -3,6 +3,8 @@
 let
   community-version = lib.versions.majorMinor python.version;
 
+  pylibs-dir = ".pythonlibs";
```
cdmistman, line 6: is this a standard directory in pythonland?

cdmistman: oh, nevermind, i see we're already using that dir

cdmistman (approved):
need to merge first

ryantm (approved):

pkgs/modules/python/default.nix
airportyh: No. I made it up.

PR #40, opened 2023-06-07
Update bun

cdmistman:
Why
===

There were [8 releases of bun in 2 weeks](https://bun.sh/blog/bun-v0.6.7) starting with 0.6.0 on May 16th. Seems this has died down a bit, so updating the module here. Some features I'm excited about:
- *bun*dler
- perf boosts
- node api catchup
- test suite improvements
- bug fixes

What changed
============

- Added `bun-0.6` module
- Added auto-upgrade to the latest version in the upgrade map
  - There weren't any announced breaking changes, so this auto-upgrade shouldn't negatively impact any Repls

Test plan
=========

`bun --version` prints the right version

Rollout
=======


ryantm (approved):
Builds locally for me, looks good. Thanks for removing the extra params!

PR #43, opened 2023-06-08
Fix node path

airportyh:
Why
===

Bug:
1. install node module in blank repl
2. node binary not found

Reason: I set the `PATH` env var in the module, overriding the `PATH` generated by lib.makeBinPath

This is not the first time I made this mistake, so to prevent future mistakes, I merged the PATHs automatically.

What changed
============

Added function to automatically merge the `PATH` variable if one is specified by the module.

Test plan
=========
```
nix build .#modules.'"nodejs-18"'
cat result | jq .env.PATH
```
See that the output has 2 entries under $REPL_HOME  and one that comes from /nix/store

Rollout
=======

### DIFF
--- a/pkgs/moduleit/module-definition.nix
+++ b/pkgs/moduleit/module-definition.nix
@@ -567,14 +567,18 @@ in
 
     replit.buildModule =
       let
-
+        envWithMergedPath = (env: path:
+          let packagesPath = lib.makeBinPath config.packages;
+          in
+          if builtins.hasAttr "PATH" env
+          then env // { PATH = env.PATH + ":" + packagesPath; }
+          else { PATH = packagesPath; } // env
+        );
         moduleJSON = {
           id = config.id;
           name = config.name;
           description = config.description;
-          env = {
-            PATH = lib.makeBinPath config.packages;
-          } // config.replit.env;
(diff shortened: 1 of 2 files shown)


pkgs/moduleit/module-definition.nix
```diff
@@ -567,14 +567,18 @@ in
 
     replit.buildModule =
       let
-
+        envWithMergedPath = (env: path:
+          let packagesPath = lib.makeBinPath config.packages;
+          in
+          if builtins.hasAttr "PATH" env
+          then env // { PATH = "${env.PATH}\:${packagesPath}"; }
```
ryantm, line 574:
```suggestion
          then env // { PATH = "${env.PATH}:${packagesPath}"; }
```
Is there a reason to escape the colon? I tried it either way and it seems the same...

ryantm (approved):

pkgs/moduleit/module-definition.nix
airportyh: Weird, it was not working for me with it. Let me try again.

ryantm: lol, fine to keep it then, it doesn't break anything.

PR #44, opened 2023-06-08
Poetry wrapper

airportyh:
Why
===

Turns out that even though we put poetry in its own venv it's still not isolated enough. Bug:

1. make blank repl
2. install python 3.10 module
3. `poetry init` walk through prompts
4. `poetry add urllib3`
5. `poetry pyyaml` fails due to

```
'HTTPResponse' object has no attribute 'strict'
```

Unsetting `PYTHONPATH` fixes it.

What changed
============

Added poetry-wrapper to freeze `PYTHONPATH` for poetry.

Test plan
=========

Re-test the bug and it should be able to install pyyaml.

Rollout
=======

- [ x] This is fully backward and forward compatible


ryantm (approved):
The unsettings will continue until morale improves.

PR #49, opened 2023-06-13
Python tweaks

airportyh:
Why
===

1. error when a library depends on setuptools' pkg_resources - when using python out of the nix store directly, setuptools isn't installed  and we get error: [No module named pkg_resources](https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources).
when using a library that depends on `pkg_resources`. One instance is https://pypi.org/project/chromadb/. The standard solution in the Nix world is to use the `buildPythonApplication` helper and put
pkgs.pythonPackages.setuptools in `propagatedBuildInputs`. But that doesn't apply to our case because we don't have the source code of the user that would comprise the application yet.
2. remove some extras from python lsp server - I am getting warnings from our the linter that don't show up in our standard python template. This is because we added all the lsp extra plugins.

What changed
============

1. append setuptool's lib path to `PYTHONPATH` for the python and prybar wrappers
2. Changed extras list to match the one in the current python template: https://replit.com/@replit/Python#pyproject.toml

Test plan
=========

1. Can run a the chroma db hello world after installing it:
    * `poetry init`
    * `poetry add chromadb`
    * [use this](https://gist.github.com/airportyh/460e77f03942b951a1688dcc36f874b5) as main.py
    * run it with python
    * run it with prybar
2. No warnings when you have no extra blank line at end of file

Rollout
=======


pkgs/python-lsp-server/default.nix
```diff
@@ -40,19 +40,20 @@ pypkgs.buildPythonPackage rec {
...
     yapf
     pyflakes
     rope
+    # autopep8
+    # flake8
+    # mccabe
+    # pycodestyle
+    # pydocstyle
+    # pyflakes
+    # pylint
```
ryantm, line 56: Are these commented out on purpose?

airportyh: I thought I leave these here so in the future we know what the available ones are.

ryantm (approved):

PR #51, opened 2023-06-14
Poetry speed up

airportyh:
Why
===

Get this poetry update https://github.com/replit/poetry/pull/5

What changed
============

Updated poetry version to 1.14.

Test plan
=========

Once this disk is in canary

1. Create new repl from https://replit.com/@replit/Python-Beta?v=1
2. `poetry add requests` should not cause any update operations to occur


ryantm (approved):
:fast_forward:

PR #53, opened 2023-06-16
Th java language server

airportyh:
Why
===

Switch to https://github.com/georgewfraser/java-language-server for less memory and cpu consumption. Get this
[configuration feature in](https://github.com/replit/java-language-server/pull/1).

What changed
============

1. Added nix file for java-language-server

Test plan
=========

Await https://github.com/replit/java-language-server/pull/1 to be merged so we can updated the revs and shas
1. Make blank repl
2. install the Java Graalvm module
3. Make a swing program `Main.java` based on https://gist.github.com/airportyh/b97570b1c46d839678ebd14e6e8e690d
4. See that you can hover over `JFrame` and get info on it

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,12 +3,14 @@
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-23.05";
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
   inputs.prybar.url = "github:replit/prybar";
+  inputs.java-language-server.url = "github:replit/java-language-server";
+  inputs.java-language-server.inputs.nixpkgs.follows = "nixpkgs";
 
-  outputs = { self, nixpkgs, nixpkgs-unstable, prybar, ... }:
+  outputs = { self, nixpkgs, nixpkgs-unstable, prybar, java-language-server, ... }:
     let
       mkPkgs = nixpkgs-spec: system: import nixpkgs-spec {
         inherit system;
-        overlays = [ self.overlays.default prybar.overlays.default ]; # ++ import ;
+        overlays = [ self.overlays.default prybar.overlays.default java-language-server.overlays.default ]; # ++ import ;
         # replbox has an unfree license
         config.allowUnfreePredicate = pkg: builtins.elem (nixpkgs.lib.getName pkg) [
           "@replit/replbox"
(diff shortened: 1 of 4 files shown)


pkgs/java-language-server/default.nix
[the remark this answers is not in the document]
airportyh: True, but I could see benefit of keeping this here too. It is modified from https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/development/tools/java/java-language-server/default.nix so we can diff it against upstream if needed.

@ryantm do you have a preference here?

flake.nix
```diff
@@ -3,12 +3,13 @@
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-23.05";
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
   inputs.prybar.url = "github:replit/prybar";
+  inputs.java-language-server.url = "github:replit/java-language-server";
```
ryantm, line 6:
```suggestion
  inputs.java-language-server.url = "github:replit/java-language-server";
  inputs.java-language-server.inputs.nixpkgs.follows = "nixpkgs";
```
Ideally, we use the same nixpkgs as everything else  or "nixpkgs-unstable" to limit the number of nixpkgs checkouts we have to evaluate.

pkgs/modules/java/default.nix
```diff
@@ -1,19 +1,31 @@
-{ pkgs, lib, ... }:
+{ pkgs-unstable, lib, ... }:
 
 let
-  graalvm = pkgs.graalvm17-ce;
+  pkgs = pkgs-unstable;
+  graalvm = pkgs.graalvm19-ce;
```
ryantm, line 5: So weird that  "19-ce" in "graalvm19-ce" is version "22.3", but that seems right:

image

pkgs/modules/java/default.nix
```diff
@@ -1,19 +1,31 @@
-{ pkgs, lib, ... }:
+{ pkgs-unstable, lib, ... }:
```
ryantm, line 1: Why do we need to use unstable?

ryantm (approved):

pkgs/modules/java/default.nix
airportyh: The stable one didn't have version 19 of graalvm, which is needed to compile the project.

airportyh: Now that we switched to 23.05 probably don't need to.

PR #57, opened 2023-07-05
added optionFileParam attribute to runners

airportyh:
Why
===

Prompted by https://replit.slack.com/archives/C03KS2B221W/p1687220639216759, we need to preserve the existing interpreter behavior. One of the behaviors that was broken with Nix modules is when you initially open a repl with an interpreter setup with prybar, you can use the "Console" pane to enter code in the language of the repl before even hitting the run button for the first time.

Plan:

1. allow a runner to optionally take a file param with the `optionalFileParam` attribute. Only one of `optionalFileParam` and `fileParam` should be set to true
2. in pid1, an optionalFileParam runner can be run without a fileParam and will be utilized at initialization time of the interp2 service via `p.freshInterp("", p, nil)`
3. the front end will use new attribute to match files to run configs

What changed
============

1. Added `optionalFileParam` to the runner options
2. In order to not propagate a change to all modules, I wrote a `stripAttrsWithDefaultValues` function that will strip out attributes from an attribute set if its value is the same as the expected default value, in this case we only use it with `optionalFileParam`. Didn't want to use it with everything, because that would again trigger changes to all modules.
3. assertion to make sure a runner does not have both `optionalFileParam` and `fileParam` set to true
4. Updated Node.js modules with `optionalFileParam` = true

Test plan
=========

1. `cd pkgs/moduleit`
2. change example.nix to have a `optionalFileParam` set and building `./moduleit.sh example.nix`
3. try having `optionalFileParam` and `fileParam` true, building should fail

Rollout
=======

### DIFF
--- a/pkgs/moduleit/module-definition.nix
+++ b/pkgs/moduleit/module-definition.nix
@@ -79,7 +79,7 @@ let
     };
   };
 
-  runnerModule = { name, config, ... }:
+  runnerOptions =
     let
       runnerProductionOverrides = { name, config, ... }:
         {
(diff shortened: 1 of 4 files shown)


ryantm (approved):

pkgs/moduleit/module-definition.nix
```diff
@@ -110,71 +110,79 @@ let
...
+      };
 
-        interpreter = mkOption {
-          type = types.bool;
-          default = false;
-          description = lib.mdDoc ''
-            Whether this runner starts an interpreter. Default: false.
-          '';
-        };
+      optionalFileParam = mkOption {
```
masad-frost, line 142: Nit: Is there a way to make these mutually exclusive?

e.g. `fileParam: { optional: true }`/`fileParam { optional: false }` or something?

airportyh: I added an assertion that makes sure you can set both of them https://github.com/replit/nixmodules/blob/b81148d6b7dfb904f9d894ebe9805e209efb430c/pkgs/moduleit/module-definition.nix#L606

masad-frost: I see, thanks

masad-frost (approved):
not sure if you're waiting for my review :shipit:  hehe

PR #65, opened 2023-07-13
[python] fewer libraries for pandas / numpy

ryantm:
Why
===
* We encountered an issue where a downloaded binary used libgcc_s.so from the stdenv.cc.cc.lib when the binary was called from a Python program. Numpy doesn't need all the libraries in stdenv.cc.cc.lib.

What changed
===
* Be more explicit about what binaries we need for numpy/pandas

Test plan
===
* Confirm we import numpy and pandas
* Confirm something using C++ parts of numpy works:

```
import numpy

arr = numpy.array([3, 2, 0, 1])

print(numpy.sort(arr, kind='heapsort'))
```

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -223,6 +223,10 @@
     "commit": "4ba5dba393f9a1dde13b73b94a58b344ca0cc9ac",
     "path": "/nix/store/sip8bp1wzdfyrx4j8ws9dpkczx2wf0b7-replit-module-python-3.10"
   },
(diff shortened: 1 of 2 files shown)


masad-frost (approved):

lhchavez (approved):
nice! is there a way of exporting that `cppLibs` to our pid1 bundle so that we can use the same trick in the Python template?

airportyh (approved):

PR #67, opened 2023-07-19
update bun package version

cdmistman:
Why
===

bun 1.0 is coming out soon, we'll wanna drum up hype

What changed
============

- update bun module to 0.7.14
- change runner to attempt using `package.json` first

Test plan
=========

- install this bun module in a Bun template repl
- `bun --version` should print `0.7.14`
- run button still works
- change `package.json#main` to be `src/index.ts`
- run button fails

Rollout
=======

### DIFF
--- a/pkgs/modules/bun/default.nix
+++ b/pkgs/modules/bun/default.nix
@@ -6,6 +6,23 @@ let
   extensions = [ ".json" ".js" ".jsx" ".ts" ".tsx" ];
 
   community-version = lib.versions.majorMinor bun.version;
+
+  package-json-runner = pkgs.writeScript "bun-runner" ''
+    #!${pkgs.bash}/bin/bash
+    if start=$(${pkgs.jq}/bin/jq -r '.scripts.start? // empty' package.json 2>/dev/null); [[ -n $start ]]; then
+      echo "+ bun run start";
+      ${bun}/bin/bun run start;
+    elif module=$(${pkgs.jq}/bin/jq -r 'if .type == "module" then .module // empty else empty end' package.json 2>/dev/null); [[ -n $module ]]; then
+      echo "+ bun $module";
+      ${bun}/bin/bun $module;
+    elif main=$(${pkgs.jq}/bin/jq -r '.main // empty' package.json 2>/dev/null); [[ -n $main ]]; then
+      echo "+ bun $main";
+      ${bun}/bin/bun $main;
+    elif [[ -n $file ]]; then
+      echo "+ bun $file";
+      ${bun}/bin/bun $file;
+    fi
+  '';
 in
 
 {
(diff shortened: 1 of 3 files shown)


pkgs/modules/bun/default.nix
```diff
@@ -6,6 +6,10 @@ let
   extensions = [ ".json" ".js" ".jsx" ".ts" ".tsx" ];
 
   community-version = lib.versions.majorMinor bun.version;
+
+  package-json-runner = pkgs.callPackage ../../package-json-runner {
+    js-runtime = "bun";
```
airportyh, line 11: Why not use the absolute path to the executable?

pkgs/package-json-runner/default.nix
```diff
@@ -0,0 +1,11 @@
+{ jq, js-runtime, writeScript }: writeScript "run-${js-runtime}" ''
+  if start=$(${jq}/bin/jq -r '.scripts.start? // empty' package.json 2>/dev/null); [[ -n $start ]]; then
+    eval $start;
```
airportyh, line 3: does bun have an equivalent to `npm run <command>`?

pkgs/modules/bun/default.nix
```diff
@@ -29,7 +33,7 @@ in
     language = "javascript";
     inherit extensions;
 
-    start = "${bun}/bin/bun run $file";
+    start = "eval ${package-json-runner}";
```
airportyh, line 36: fileParam should probably be changed to `optionalFileParam` because the package-json-runner can run without a file parameter as well.

pkgs/modules/bun/default.nix
cdmistman: to get the nice name in the runner's derivation name 😅 wasn't sure if i'd be bringing it back or not - it's coming back in one of the next commits :)

pkgs/package-json-runner/default.nix
cdmistman: yes! good point lmao why would i do this

cdmistman: ohhhhhhhh actually i see what's going on here. ok i'm not gonna keep trying this package-json-runner abstraction, that can happen later

airportyh (approved):
Approve for now to unblock. But I think you should change fileParam -> optionalFileParam before merging.

cdmistman:
> should change fileParam -> optionalFileParam before merging

`fileParam` should probably accept `true`, `false`, or `optional`

airportyh (approved):

airportyh:
> > should change fileParam -> optionalFileParam before merging
> 
> `fileParam` should probably accept `true`, `false`, or `optional`

I didn't want to change the API which would cause every module to change.

PR #70, opened 2023-07-24
SvelteKit

cdmistman:
Why
===

A really popular (and getting more popular) frontend framework.

What changed
============

- added svelte-language-server to the nodejs module

Test plan
=========

- open a `.svelte` file
- type `<svelte:he` and notice completion for `<svelte:head`
- profit

Rollout
=======

### DIFF
--- a/pkgs/modules/default.nix
+++ b/pkgs/modules/default.nix
@@ -11,12 +11,14 @@ let
       pypkgs = pkgs.python310Packages;
     }))
     (mkModule ./pyright-extended)
+
     (mkModule (import ./nodejs {
       nodejs = pkgs.nodejs-18_x;
     }))
     (mkModule (import ./nodejs {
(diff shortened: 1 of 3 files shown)


pkgs/modules/default.nix
```diff
@@ -11,12 +11,18 @@ let
...
     (mkModule ./pyright-extended)
+
     (mkModule (import ./nodejs {
       nodejs = pkgs.nodejs-18_x;
     }))
     (mkModule (import ./nodejs {
       nodejs = pkgs.nodejs_20;
     }))
+
+    (mkModule (import ./svelte-kit {
```
airportyh, line 22: You should be able to do just `(mkModule ./svelte-kit)` and it will have access to pkgs-unstable. Look at the rust module.

airportyh (approved):

PR #72, opened 2023-07-25
Added bundle-squashfs-fn and bundle-locked-fn to allow customization …

airportyh:
…of what modules to build into a disk

Why
===

Building the entire Nix modules disk is time consuming and often requires more free disk space than my computer has.
We want to make it easy to build a disk with a customized set of modules so testing a new module change can go faster.

What changed
============

Added functions: `bundle-squashfs-fn` and `bundle-locked-fn` which allow you to pass a list of module IDs. See the docs at the top of bundle-locked/default.nix for the exact allowable formats and what they mean.

Test plan
=========

try calling `nix build .#custom-bundle-locked` and `nix build .#custom-bundle-squashfs` and also changing the list of module IDs that go into them.


pkgs/default.nix
```diff
@@ -28,18 +28,39 @@ rec {
...
+    moduleIds = null;
+  };
+
+  basic-bundle-locked = bundle-locked-fn {
+    moduleIds = [ "python-3.10:v1-20230525-c48c43c" ];
+  };
+
+  basic-bundle-squashfs = bundle-squashfs-fn {
+    moduleIds = [ "python-3.10" "nodejs-18" ];
   };
```
cdmistman, line 64: can we just generate an output for every module instead of this?

airportyh: @ryantm didn't like there being so many entries in `nix flake show`? Maybe we can nest in under another attrset like modules.

cdmistman: hm, it's just weird to me having these outputs pre-defined and maybe arbitrarily changing in the future. maybe comment these out and leave a comment saying 'these are how you test the modules locally'?

airportyh: I'll document a good flow...

airportyh: This is what I have it mind: https://github.com/replit/goval/pull/10077

pkgs/default.nix
```diff
@@ -22,24 +23,58 @@ rec {
...
-  bundle-squashfs = pkgs.callPackage ./bundle-squashfs {
-    inherit bundle-locked revstring;
-    inherit active-modules upgrade-maps;
+  bundle-locked-fn = { modulesLocks }: pkgs.callPackage ./bundle-locked {
+    inherit modulesLocks;
+    inherit revstring;
+  };
+
+  bundle-locked = bundle-locked-fn {
+    moduleIds = null;
```
ryantm, line 46: how does this work? the bundle-locked-fn says it wants 1 paramater called modulesLocks but this is passing moduleIds

airportyh: oops.

pkgs/default.nix
```diff
@@ -22,24 +23,58 @@ rec {
...
     inherit bundle-locked revstring;
     inherit active-modules upgrade-maps;
   };
 
   bundle-image-tarball = pkgs.callPackage ./bundle-image-tarball { inherit bundle-image revstring; };
 
-  bundle-squashfs = pkgs.callPackage ./bundle-squashfs {
-    inherit bundle-locked revstring;
-    inherit active-modules upgrade-maps;
+  bundle-locked-fn = { modulesLocks }: pkgs.callPackage ./bundle-locked {
```
ryantm, line 40: This and the other -fn break `nix flake show`, we either need to move them up into the let block at the top of this file or make a new lib output for the flake and then reference them via that.

ryantm (changes requested):

ryantm (approved):

PR #74, opened 2023-07-28
remove TS LSP from SvelteKit

cdmistman:
Why
===

Turns out Svelte's LSP already has JS/TS file support

What changed
============

removed TS LSP from SvelteKit

Test plan
=========

- open foo.js
- enter invalid expression
- only svelte LSP should report the error

Rollout
=======


pkgs/default.nix
```diff
@@ -79,7 +79,7 @@ rec {
     # customize these IDs for dev. They can be like "python-3.10:v10-20230711-6807d41" or "python-3.10"
     # publish your feature branch first and make sure modules.json is current, then
     # in goval dir (next to nixmodules), run `make custom-nixmodules-disk` to use this disk in conman
-    moduleIds = [ "python-3.10" "nodejs-18" ];
+    moduleIds = [ "svelte-kit-node-20" ];
```
ryantm, line 82: Let's not commit this part

ryantm (changes requested):

ryantm (approved):

PR #75, opened 2023-08-02
phony OCI image bundle

ryantm:
Why
===
* We'd like to be able to quickly build containers that depend on nixmodules.

What changed
===
* Add new bundle-phony-ocis package output that makes a phony OCI image for each nixmodule. The OCI images contain a layer that is the /nix/store paths of the module, and a ZTOC blob for that layer. These modules cna be uploaded to a Registry.

Test plan
===
* built `nix build .#bundle-phony-ocis` output looks reasonable
* can push output to Registry

Rollout
=======

### DIFF
--- a/pkgs/mk-phony-oci/builder.sh
+++ b/pkgs/mk-phony-oci/builder.sh
@@ -1,0 +1,81 @@
+set -eou pipefail
+. .attrs.sh
+
+PATH="${env[PATH]}"
+out="${outputs[out]}"
+mkdir $out
+
+mkdir image
+echo '{"imageLayoutVersion": "1.0.0"}' > ${out}/oci-layout
+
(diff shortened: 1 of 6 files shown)


pkgs/mk-phony-oci/builder.sh
```diff
@@ -0,0 +1,81 @@
+set -eou pipefail
+. .attrs.sh
```
cdmistman, line 2: where's this file?

cdmistman: oh nevermind, it's a stdenv file

ryantm: It gets created via `__structuredAttrs = true;`

cdmistman (approved):
no way to test as i deleted my devvm but LGTM

PR #76, opened 2023-08-03
Upgrade poetry to 1.5.1

airportyh:
Why
===

Upgrade to the latest stable 1.5 line of poetry. We still have [customizations](https://github.com/replit/poetry) but they are current to the [upstream line](https://github.com/python-poetry/poetry/tree/1.5).

What changed
============

Updated poetry version and added new customization flags.

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -209,7 +209,14 @@ in
       PYTHONPATH = "${python}/lib/python${community-version}:${userbase}/${python.sitePackages}";
       PIP_CONFIG_FILE = pip-config.outPath;
       POETRY_CONFIG_DIR = poetry-config.outPath;
+      POETRY_CACHE_DIR = "$REPL_HOME/.cache/pypoetry";
       POETRY_VIRTUALENVS_CREATE = "0";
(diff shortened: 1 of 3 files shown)


pkgs/modules/python/default.nix
```diff
@@ -209,7 +209,14 @@ in
...
       PIP_CONFIG_FILE = pip-config.outPath;
       POETRY_CONFIG_DIR = poetry-config.outPath;
+      POETRY_CACHE_DIR = "$REPL_HOME/.cache/pypoetry";
       POETRY_VIRTUALENVS_CREATE = "0";
+      POETRY_INSTALLER_MODERN_INSTALLATION = "0";
+      POETRY_PIP_USE_PIP_CACHE = "1";
+      POETRY_PIP_NO_ISOLATE = "1";
+      POETRY_PIP_NO_PREFIX = "1";
+      POETRY_PIP_FROM_PATH = "1";
+      POETRY_USE_USER_SITE = "1";
```
ryantm, line 219: Is this the place to put these, or should we do in the poetry wrapper above?

airportyh: Decided to do it here so that we could overwrite these in .replit for testing if needed.

ryantm (approved):

PR #80, opened 2023-08-15
add gcloud module

ryantm:
Why
===
* We'd like to make it very easy to install Google Cloud tools

What changed
===
* Add module that installs google-cloud-sdk

Test plan
===
* builds fine locally, can run binaries from result

Rollout
=======


airportyh (approved):
Thanks!

phil-mac:
Is there a deployment process that's needed for this new `gcloud` nixmodule to be available? I tried `await toolchain.addModule('gcloud')` and I see `modules = ["gcloud"]` in `.replit` but gcloud commands in the shell return `gcloud: command not installed, but was located via Nix.`

PR #84, opened 2023-08-23
use /repl/stats/resources.json for memory limit

cdmistman:
Why
===

we want to enable cgroupsv2 in replspace but JDT Language Server has a hard dependency on cgroups v1

What changed
============

use `/repl/stats/resources.json` instead of the cgroups file path

Test plan
=========

1. boot repl with Java nixmodule
2. Java LSP still works
3. enable cgroupsv2 in replspace
4. Java LSP still works

Rollout
=======


lhchavez (approved):

ryantm (approved):
Thanks!

PR #85, opened 2023-08-24
[python] add escape hatch for LD_LIBRARY_PATH

ryantm:
Why
===
* Currently there is no way for the end user to add dynamic libraries to the Python environment

What changed
===
* Introduce PYTHON_LD_LIBRARY_PATH (an env var we had in older Replit Python templates). If this variable is set in the environment, use it for Python's LD_LIBRARY_PATH and ignore the default settings. If someone uses the escape hatch, they have to go all the way.
* Make a pythonWrapper helper function that allows us to keep all the wrappers consistent

Test plan
===
* Ran all the wrapped binaries. Tested that LD_LIBRARY_PATH is what I expect.

Rollout
=======


airportyh (approved):
Nice extraction!

airportyh (approved):

PR #86, opened 2023-08-25
use jq directly from pkgs

cdmistman:
Why
===

Made an oopsie here too

What changed
============

fixed the jq call for jdt-language-server

Test plan
=========

jdt language server works

Rollout
=======


ryantm (commented):
Thanks, sorry I missed it last time.

ryantm (approved):

PR #87, opened 2023-08-25
python 3.11

airportyh:
Why
===

Want to add Python 3.11 module.

What changed
============

1. because debugpy has known bugs on Python 3.11, disabled it for python-3.11.
2. Built and referenced new Poetry bundle for python 3.11.
3. Refactored poetry-in-venv setup.
4. renamed community-version to PythonVersion because we don't really call it that anymore

Test plan
=========


ryantm (approved):
Looks good

PR #90, opened 2023-09-03
[WS-967] Add prettier to node nixmodules

bradymadden97:
Why
===

https://linear.app/replit/issue/WS-967/add-prettier-to-jsts-nixmodules

Trying to support prettier as a more built-in experience in js/ts repls. The idea here is:
* Add it to nix modules
* Expose it as a `formatters` through toolchain config
* Allow user to select it as their formatter (or use lsp formatter, etc)

What changed
============

* Following Toby's high level steps here to add prettier and expose through `formatters` field
* Ran `python scripts/lock_modules.py`

Test plan
=========

I'm not sure how to test this. If there's a way to point a local replit web at this so I can confirm prettier is infact installed, that sounds good. But I might need some help figuring out how to do that.

Rollout
=======

### DIFF
--- a/pkgs/modules/nodejs/default.nix
+++ b/pkgs/modules/nodejs/default.nix
@@ -15,6 +15,8 @@ let
     ${prybar}/bin/prybar-nodejs -q --ps1 "''$(printf '\u0001\u001b[33m\u0002\u0001\u001b[00m\u0002 ')" -i ''$1
   '';
 
+  prettier = nodepkgs.prettier;
+
 in
 
 {
@@ -28,6 +30,7 @@ in
 
   packages = [
     nodejs
+    prettier
   ];
 
   replit = {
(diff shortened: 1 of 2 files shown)


pkgs/modules/nodejs/default.nix
```diff
@@ -15,6 +15,8 @@ let
     ${prybar}/bin/prybar-nodejs -q --ps1 "''$(printf '\u0001\u001b[33m\u0002\u0001\u001b[00m\u0002 ')" -i ''$1
   '';
 
+  prettier = pkgs.nodePackages.prettier;
```
airportyh, line 18: Usually I prefer to pin the node.js used by the tools to the same version as supplied by the module. Using `nodepkgs` instead of `pkgs.nodePackages` does this.
```suggestion
  prettier = nodepkgs.prettier;
```

airportyh (changes requested):
Just the nodepkgs change.

airportyh (changes requested):
Sorry, would you mind manually squashing the commits into 2 to reduce the commit log in main?

This is what I do:
```
git fetch
git merge origin/main # just in case
git reset origin/main
git checkout modules.json
git add .
git commit -m "<you commit message>"
python3 scripts/lock_modules.py
git commit -m "updated node.js modules.json"
git push --force
```

Sorry for the inconvinience.

airportyh (approved):

PR #92, opened 2023-09-05
[ci] try using Magic Nix Cache

ryantm:
Why
===
* We don't want to have it build everything every time
* https://github.com/DeterminateSystems/magic-nix-cache-action

What changed
===
* Use Determinate Systems installer and magic nix cache in CI workflow

Test plan
===
* CI passes

Rollout
=======


airportyh (approved):
Nice!

PR #93, opened 2023-09-06
feat: add nix module with nixd language server

jakehamilton:
Why
===

We don't currently have a language server running for Nix files which can make it feel like you're flying blind. Instead, we can leverage [`nixd`](https://github.com/nix-community/nixd) to provide a rich experience and potentially options completion in the future.

What changed
============

We now pull in `nixd` via flake input and a new `nix` module was added that configures the language server for Nix files.

Test plan
=========

Help! What is a good way for me to test that this works?

Rollout
=======

I don't believe this will introduce any incompatibilities. Should be good to roll out once confirmed working and any feedback is addressed.

### DIFF
--- a/pkgs/modules/nix/default.nix
+++ b/pkgs/modules/nix/default.nix
@@ -1,0 +1,15 @@
+{ pkgs, ... }: {
+  id = "nix";
+  name = "Nix";
+
+  replit.languageServers.nixd = {
+    name = "nixd";
+    language = "nix";
+    start = "${pkgs.nixd}/bin/nixd";
+    extensions = [ ".nix" ];
+
+    configuration = {
+      options.enable = false;
+    };
+  };
+}
--- a/flake.lock
+++ b/flake.lock
@@ -56,6 +56,24 @@
         "type": "github"
       }
     },
+    "flake-parts": {
+      "inputs": {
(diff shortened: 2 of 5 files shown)


pkgs/modules/nix/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  name = "Nix";
+
+  replit.languageServers.nixd = {
+    name = "nixd";
+    language = "nix";
+    start = "${pkgs.nixd}/bin/nixd";
+    extensions = [ ".nix" ];
+
+    configuration = {
+      options.enable = false;
```
ryantm, line 12: What's this part for? I would think we'd want to enable it by default since the way we enable modules right now is by installing them.

jakehamilton: The language server can provide completion for NixOS modules. I figured we wanted to turn that specific part off right now since we don't want/need completions for things like `config.services.*`.

It is configurable, though and we could enable completion for Replit config in the future!

jakehamilton: See this section for more info:

https://github.com/nix-community/nixd/blob/main/docs/user-guide.md#options

masad-frost (commented):
Meta

ryantm (approved):

PR #94, opened 2023-09-07
[WS-967] Update prettier start command

bradymadden97:
Why
===

When the start command is a string, it gets modified into `sh -c <command>`. See https://github.com/replit/goval/pull/10626#discussion_r1317457548

We don't actually want `sh -c`, because we need to pipe the output from stdin, so let's pass the start command in a different format.

What changed
============

Use an array of args for start command

Test plan
=========

Tested this end to end using `custom-nixmodules-disk`, loaded it in local conman, and was able to run prettier in local repl-it-web

Rollout
=======


ryantm (approved):
Thanks, looks good and looks like you're learning a lot about nixmodules too!

PR #99, opened 2023-09-13
use internal bun drv and bun 1.0.1

cdmistman:
Why
===

been getting a lot of bumps about updating bun package for replit, but relying on upstream nixpkgs for this means we actually lose 2 workdays (not including weekends) every time there's a release before we're caught up on bun versions

we'll just internally maintain a duplicate version of the bun derivation from upstream, but once the hype dies down we'll want to remove it and rely on upstream

What changed
============

- brought the bun derivation file into this repo
- use the internal derivation for bun module
- update bun to 1.0.1

Test plan
=========

- install `bun-1.0` module v2
- `bun --version` prints `1.0.1`

Rollout
=======


pkgs/upgrade-maps/default.nix
```diff
@@ -16,6 +16,7 @@ let
     "bun-0.6:v3-20230623-0b7a606" = { to = "bun-0.6:v4-20230721-719ce58"; auto = true; };
     "bun-0.6:v4-20230721-719ce58" = { to = "bun-0.7:v1-20230724-4274858"; auto = true; };
     "bun-0.7:v1-20230724-4274858" = { to = "bun-1.0:v1-20230911-f253fb1"; auto = true; };
+    "bun-1.0:v1-20230911-f253fb1" = { to = "bun-1.0:v1-20230911-f253fb1"; auto = true; };
```
relevantsam, line 19: Should this be to v2?

cdmistman: ope good catch thanks

pkgs/modules/bun/bun.nix
```diff
@@ -0,0 +1,82 @@
...
+      All in one fast & easy-to-use tool. Instead of 1,000 node_modules for development, you only need bun.
+    '';
+    license = with licenses; [
+      mit # bun core
+      lgpl21Only # javascriptcore and webkit
+    ];
+    maintainers = with maintainers; [ DAlperin jk thilobillerbeck cdmistman coffeeispower ];
+    platforms = builtins.attrNames passthru.sources;
+  };
+}
```
ryantm, line 82: Did you consider using overrides instead? It seems to work:
```suggestion
{ bun
, fetchurl
}:

bun.overrideAttrs rec {
  version = "1.0.1";
  src = fetchurl {
    url = "https://github.com/oven-sh/bun/releases/download/bun-v${version}/bun-linux-x64.zip";
    hash = "sha256-RmgnWTG6kTebYwIa/VAwvvJmbL+ARNC+HkbF4mJPF7o=";
  };
}
```

ryantm (approved):

PR #100, opened 2023-09-13
Remove unnecessary constraint on debugpy

noneofyourbusiness1415252:
Why
===

https://ask.replit.com/t/python-repls-do-not-offer-a-debugger-tool-after-nix-conversion/63323

What changed
============

Remove check for whether the `python` argument passed to the `python` module is versioned 3.8.

Test plan
=========
can't figure out how to (`nix build`'s trying to `chmod` `/nix/store/` paths) :shrug:. But `python38Packages.debugpy` does exist on nix 22.11

Rollout
=======

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,11 @@
+entrypoint = "main.py"
+modules = ["python-3.10:v18-20230807-322e88b"]
+
+hidden = [".pythonlibs"]
+
+[nix]
+channel = "stable-22_11"
+[env]
+NIX_CONFIG = "extra-experimental-features = nix-command flakes"
+[deployment]
+run = ["sh", "-c", "python3 main.py"]
--- a/modules.json
+++ b/modules.json
@@ -435,6 +435,10 @@
     "commit": "1095880875b051ac1690ea27bc76803db516f800",
     "path": "/nix/store/yq9na7hmkrjhr2410prm1a7410nwbhnw-replit-module-python-3.8"
   },
+  "python-3.8:v4-20230920-8e98cf2": {
+    "commit": "8e98cf2cd862f686c872ad83d5f51a5f04d0de76",
+    "path": "/nix/store/3ssbfv5kig6rw6q8rsrabkjf2aqqdvz2-replit-module-python-3.8"
+  },
   "bun-1.0:v1-20230911-f253fb1": {
     "commit": "f253fb1a97d9a1c950d90081ee1ccb290776cf8b",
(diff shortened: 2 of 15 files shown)


noneofyourbusiness1415252:
update version constraint to 3.13 as Nix is doing

.replit
cdmistman, line 1: please delete this - currently this repo doesn't work on replit itself

replit.nix
cdmistman, line 1: ditto - please delete

pkgs/modules/python/default.nix
```diff
@@ -98,7 +98,7 @@ let
     inherit pkgs python pypkgs debugpy;
   };
 
-  debuggerConfig = if pythonVersion == "3.8" then ({ }) else
+  debuggerConfig = if pythonVersion == "3.13" then ({ }) else
```
cdmistman, line 101: no need to check for 3.13 - we'll take care of that part

modules.json
cdmistman, line 1: please revert this file to what it was before you opened this branch - we'll do a rebase and lock the module before merging

cdmistman (changes requested):
👋 hey Umar - thank you for contributing!

when i removed debugpy from 3.8 i remember it being a cause of some breakage, but that might've been fixed elsewhere. once you do the requested changes then i'll test it :)

bonus points if you know how to squash the commits so there's only 1 commit :D (but no worries if you don't)

.replit
noneofyourbusiness1415252: ack. I'd noticed that when the build tried to `chmod` restricted paths, so I'd been using the Shell to manually add and commit to avoid pushing these, but then I used the UI to 'resolve' a merge conflict and these were pulled in

pkgs/modules/python/default.nix
noneofyourbusiness1415252: isn't it best to check for 3.13 from now, for future-proofing? 3.13 isn't supported by `debugpy` but 3.12 is (to practically the same extent as 3.11): https://github.com/microsoft/debugpy/pull/1359

cdmistman: we don't support 3.13 yet - when we do, 3.13 might be supported by debugpy already

airportyh:
This one should work: https://github.com/replit/nixmodules/pull/120

airportyh:
Superceded by #120. Debugger for Python 3.8 has been deployed. Should work in ~1-2 hours for all users.

PR #102, opened 2023-09-15
php: remove v1 module

ryantm:
Why
===
* Hydra cannot built it because the fixed-output-derivation is not reproducible

What changed
===
* Removed php-8.1:v1 from module lock file
* Added missing upgrade map for php from v1 to v2

Test plan
===
* CI passes
* Add the old PHP nixmodule string and confirm it still installs the v2
* Check hydra builds to see if this is no longer a problem

Rollout
=======


airportyh (approved):

cdmistman (approved):
🪓🔥

PR #103, opened 2023-09-15
package.json runner 2.0

cdmistman:
Why
===

the current one only applies to bun, and some of its semantics are wrong

What changed
============

The order of precedence is now:
- `scripts.replit-dev` in `package.json`
- `scripts.dev` in `package.json`
- `main` in `package.json`
- `$file`

note:
- i removed the rule running `module`
  - oftentimes this is defined as a library module instead of a bin module
- i added `replit-dev`
  - developers often use `127.0.0.1` for local dev, but that doesn't work on replit, yet `0.0.0.0` is a theoretical security concern

Test plan
=========

in a project with the following `package.json` with any of the modules affected by this PR:
```json
{
	"main": "index.js",
	"scripts": {
		"start": "echo start",
		"dev": "echo dev",
		"replit-dev": "echo replit-dev"
	}
}
```

and the following `index.js`:
```js
console.log('index.js')
```

and the following `entrypoint.js`:
```js
console.log('entrypoint')
```

with `entrypoint = "entrypoint.js"` in `.replit` file

1. default run prints "replit-dev"
1. delete replit-dev script, run prints "dev"
1. delete dev script, run prints "index.js"
1. delete main, run prints "entrypoint"
1. delete entrypoint in `.replit`, run prints "Nothing to run." and exits with status code 1

Rollout
=======


pkgs/modules/run-package-json/default.nix
```diff
@@ -0,0 +1,19 @@
...
+
+  script = pkgs-unstable.writeScript "package-json-runner" ''
+    ${bun}/bin/bun ${./script.js}
+  '';
+in
+
+{
+  replit.runners."package.json" = {
+    name = "package.json";
+    start = "${script}/bin/package-json-runner --run-script ${runFileScript} --run-package-json-script ${runPackageJsonScript}";
```
ryantm, line 16: What about this instead (might need some slight tweaking)?
```suggestion
    start = "${bun}/bin/bun ${./script.js} --run-script ${runFileScript} --run-package-json-script ${runPackageJsonScript}";
```

ryantm (approved):

pkgs/modules/run-package-json/default.nix
cdmistman: 🤦 that should be just `${script}`. but invoking the script like in your suggestion will make users more scared to change their run configs - i'm trying to avoid that with the 2nd script. lemme try the shebang again

cdmistman: should be better now?

ryantm: Oh yeah, I hadn't thought about that showing up in the Console. Looks fine.

PR #105, opened 2023-09-15
bun + sveltekit = <3

cdmistman:
Why
===

i like bun

i like sveltekit

i like bun and sveltekit

they're better together <3

What changed
============

added new `bun-sveltekit` module

Test plan
=========

- install in sveltekit template
- things still work

Rollout
=======

### DIFF
--- a/pkgs/modules/svelte-kit/bun.nix
+++ b/pkgs/modules/svelte-kit/bun.nix
@@ -1,0 +1,41 @@
+{ pkgs-unstable, lib, ... }:
+
+let
+  bun = pkgs-unstable.callPackage ../../bun { };
+
+  community-version = lib.versions.majorMinor bun.version;
+in
+
+{
+  id = "svelte-kit-bun-${community-version}";
+  name = "SvelteKit with Bun ${community-version} Tools";
+
+  imports = [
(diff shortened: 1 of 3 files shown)


ryantm (approved):

pkgs/modules/svelte-kit/bun.nix
```diff
@@ -0,0 +1,41 @@
...
+        ".js"
+        ".ts"
+      ];
+    };
+
+    languageServers.svelte-language-server = {
+      name = "Svelte Language Server";
+      language = "svelte";
+      extensions = [ ".svelte" ".js" ".ts" ];
+      start = "${pkgs-unstable.nodePackages.svelte-language-server}/bin/svelteserver --stdio";
```
masad-frost, line 38: We shouldn't have this as a module, sveltekit should be installed in user's directory, not pulled from nix. I think more often than not, frameworks don't belong to this repo.

masad-frost (changes requested):

masad-frost:
we should lean on `bun create` in this case and templates

masad-frost (approved):
I misread what this is doing. :shipit:

PR #107, opened 2023-09-18
Added set of all oci bundles

airportyh:
Why
===

The nixmodules_oci.sh script builds an OCI for every historical module (there are ~110 currently). This is very time and disk space consuming. We can check with the docker repository to be able to only build the ones that do not already exist.

What changed
============

1. Added a mapping module ID -> OCI derivation for every historical module ID
2. Added a way to get all historical modules for scripting purposes

Test plan
=========

1. `nix eval .#all-modules --json |jq`
2. `nix build .#'all-phony-oci-bundles."nodejs-18:v10-20230914-1095880"'`

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -96,6 +96,20 @@ rec {
 
   custom-bundle-phony-ocis = mkPhonyOCIs { moduleIds = [ "nodejs-18" ]; };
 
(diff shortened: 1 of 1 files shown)


ryantm (approved):
Looks good and it works locally for me.

PR #109, opened 2023-09-20
Remove prybar

airportyh:
Why
===

We are removing prybar:
https://replit.slack.com/archives/C03KS2B221W/p1694790765815429

What changed
============

Removed it from the Node and Python modules.

Test plan
=========

1. make a node or python repl
2. click run
3. it should work not be running prybar anymore: no more code eval

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -19,7 +19,7 @@
     "crane": {
       "inputs": {
         "flake-compat": "flake-compat",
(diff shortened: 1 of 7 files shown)


ryantm (changes requested):
Can we remove it from flake.nix too?

ryantm (approved):

PR #113, opened 2023-09-20
fix package json runner

cdmistman:
Why
===

looks like bun tried to `ls /nix/store` since the package-json-runner script was a direct child of that dir

also, the Stop button wasn't working

What changed
============

- use `writeScriptBin` instead of `writeScript` for the package-json-runner script
- use `exec` instead of `spawn` to launch the command

Test plan
=========

- open bun repl
- `strace -s512 -fostrace.txt package-json-runner --run-script exit --run-package-json-script exit` doesn't yield an `openat(..., "/nix/store/", ...)` in `strace.txt`

Rollout
=======

### DIFF
--- a/pkgs/modules/run-package-json/script.js
+++ b/pkgs/modules/run-package-json/script.js
@@ -1,6 +1,6 @@
-const fs = require("node:fs");
-const path = require("node:path");
-const { spawn } = require("node:child_process");
+const fs = require("fs");
+const path = require("path");
+const { spawn, exec } = require("child_process");
 
 const args = process.argv.slice(1);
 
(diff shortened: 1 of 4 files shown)


pkgs/modules/run-package-json/script.js
```diff
@@ -1,6 +1,6 @@
-const fs = require("node:fs");
-const path = require("node:path");
-const { spawn } = require("node:child_process");
+const fs = require("fs");
+const path = require("path");
+const { spawn, exec } = require("child_process");
```
ryantm, line 3:
```suggestion
const { exec } = require("child_process");
```

ryantm (approved):
Have you tested this works locally?

cdmistman:
> Have you tested this works locally?

working on it

cdmistman:
@ryantm it works after changing the execution to node :'(

PR #114, opened 2023-09-20
linearUpgrade function, stepUpgrade function and update the maps

airportyh:
Why
===

The upgrade maps are getting annoying to maintain by hand.

What changed
============

1. Added 2 functions: `stepUpgrade` and `linearUpgrade` to declaratively generate upgrade entries.
2. Rewrote most of the upgrade maps with these.
 
Test plan
=========

Make sure we don't regress.


cdmistman:
are we not caring about changelogs anymore?

airportyh:
> are we not caring about changelogs anymore?

How do you want to see changelogs handled?

cdmistman:
i'm not sure - would be good to surface them in the eventual modules pane. but it'd be nice to have the history of the module there, and perhaps we can toast for now when the resolved module changes?

i'm already using the changelog field - do you think i should remove my changelogs?

airportyh:
> i'm not sure - would be good to surface them in the eventual modules pane. but it'd be nice to have the history of the module there, and perhaps we can toast for now when the resolved module changes?
> 
> i'm already using the changelog field - do you think i should remove my changelogs?

No you shouldn't remove them. My pr is only meant to be a refactor of existing code. I didn't change the bun stuff because it's harder to generalize, but that's total valid. But argument can be made that the auto-gen functions should be more friendly to adding change logs.

cdmistman (approved):

PR #117, opened 2023-09-22
Prybar is back: but no interp

airportyh:
Why
===

We had some minor [customer outrage](https://ask.replit.com/t/unable-to-call-python-functions-from-console/66268/15) regarding the removal of Prybar. We are going to add a Python with Prybar 3.10 module with limited support. But it will have `interpreter=false` which will cause it to stop depending on the interp2 service. We plan to deprecate and remove interp2 and shellrun2 at some point. 

What changed
============

1. Added back prybar flake
2. Extracted some common python things to python-utils
3. Created python-with-prybar-3.10 module

Test plan
=========

Test this in a Repl, prybar should be back!

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -19,7 +19,7 @@
     "crane": {
       "inputs": {
         "flake-compat": "flake-compat",
-        "flake-utils": "flake-utils",
+        "flake-utils": "flake-utils_2",
         "nixpkgs": [
           "ztoc-rs",
           "nixpkgs"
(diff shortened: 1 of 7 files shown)


ryantm (approved):
I'd much rather we go the route of pointing at an old module, but if you feel this is best okay.

airportyh:
> I'd much rather we go the route of pointing at an old module, but if you feel this is best okay.

In order to change interpreter to false, we need to make a change to the module.

PR #124, opened 2023-09-30
[Prodvana] wait before approval

pacharrin:
Why
===

Conman takes a bit to update the nixmodules after we deploy the new config; this confused the team during a recent incident where it was hard to know what version they were approving.

What changed
============

This config introduces a waiting period after deploying to Canary to wait for the change to propagate and to introduce a place where we can check for alerts in datadog,

If there is an alert, the deployment will fail and not advance to other release channels.

Test plan
=========

CD should have a new protection after deploying to Canary; if no dd alert in datadog is fired for 5 minutes, it will ask for approval.

Rollout
=======

- [X] This is fully backward and forward-compatible

### DIFF
--- a/manifests/application.pvn.yaml
+++ b/manifests/application.pvn.yaml
@@ -10,6 +10,25 @@ application:
         - name: config-raw
           runtime: marine-cycle-160323--govalctl-config-raw
           type: EXTENSION
+      protections:
+        - ref:
+            name: check-no-datadog-alerts
+            parameters:
+              - name: apiKey
+                secret:
+                  secretRef:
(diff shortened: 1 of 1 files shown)


manifests/application.pvn.yaml
```diff
@@ -10,6 +10,25 @@ application:
         - name: config-raw
           runtime: marine-cycle-160323--govalctl-config-raw
           type: EXTENSION
+      protections:
+        - ref:
+            name: check-no-datadog-alerts
```
lhchavez, line 15: v pog

lhchavez (approved):

PR #125, opened 2023-10-02
bun 1.0.3

cdmistman:
Why
===

like recent bun

What changed
============

- update bun to 1.0.3

Test plan
=========

- `bun --version` with module installed prints 1.0.3

Rollout
=======

### DIFF
--- a/pkgs/modules/bun/default.nix
+++ b/pkgs/modules/bun/default.nix
@@ -6,6 +6,23 @@ let
   extensions = [ ".json" ".js" ".jsx" ".ts" ".tsx" ];
 
   community-version = lib.versions.majorMinor bun.version;
+
+  package-json-runner = pkgs.writeScript "bun-runner" ''
+    #!${pkgs.bash}/bin/bash
+    if start=$(${pkgs.jq}/bin/jq -r '.scripts.start? // empty' package.json 2>/dev/null); [[ -n $start ]]; then
+      echo "+ bun run start";
+      ${bun}/bin/bun run start;
+    elif module=$(${pkgs.jq}/bin/jq -r 'if .type == "module" then .module // empty else empty end' package.json 2>/dev/null); [[ -n $module ]]; then
+      echo "+ bun $module";
+      ${bun}/bin/bun $module;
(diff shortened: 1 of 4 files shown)


ryantm (changes requested):
We haven't fixed the runner yet, so please do not merge this.

cdmistman:
[13b1054](https://github.com/replit/nixmodules/pull/125/commits/13b105484879a2c8101ee2c4ff0e4c682529aee7) reverted the module config to use the package-json-runner script/config from v3. the only delta is the path for the bun derivation

pkgs/modules/bun/default.nix
```diff
@@ -6,17 +6,30 @@ let
...
+    fi
+  '';
 in
 
 {
   id = "bun-${community-version}";
   name = "Bun Tools";
 
   imports = [
-    (import ../run-package-json {
```
ryantm, line 16: Are you able to delete this directory now too?

pkgs/modules/bun/default.nix
```diff
@@ -6,17 +6,30 @@ let
...
+    elif module=$(${pkgs.jq}/bin/jq -r 'if .type == "module" then .module // empty else empty end' package.json 2>/dev/null); [[ -n $module ]]; then
+      echo "+ bun $module";
+      ${bun}/bin/bun $module;
+    elif main=$(${pkgs.jq}/bin/jq -r '.main // empty' package.json 2>/dev/null); [[ -n $main ]]; then
+      echo "+ bun $main";
+      ${bun}/bin/bun $main;
+    elif [[ -n $file ]]; then
+      echo "+ bun $file";
+      ${bun}/bin/bun $file;
+    fi
```
ryantm, line 24: What about the `replit-dev` runner from before?

pkgs/modules/bun/default.nix
cdmistman: i still plan on fixing `run-package-json`, i just want to do that in a different pr. the script in this pr is the v3 runner script

ryantm (approved):

PR #129, opened 2023-10-06
Fix owner group ids

Mic92:
Why
===

Create images with owner 11000 / group 11000 so that they do not come up as the nobody user in the repl.it.

What changed
============

We switched to an lkl fork that supports this feature. An upstream pull request has been opened: https://github.com/lkl/linux/pull/532

Test plan
=========

```
$ nix build .#bundle-image
$ sudo mount -o ro,loop,norecovery ./result/disk.raw /mnt
$ ls -lan /mnt
total 37
drwxr-xr-x  5     0     0  4096 Feb  2  1970 .
drwxr-xr-x 21     0     0    21 Oct  6 13:08 ..
drwxr-xr-x  3 11000 11000  4096 Oct  6 14:24 etc
drwx------  2     0     0 16384 Oct  6 14:29 lost+found
drwxr-xr-x  3 11000 11000  4096 Feb  2  1970 nix
```

Rollout
=======

- [x] This is fully backward and forward compatible (assuming that no one depends on quirky behavior where files are owned by nobody)

### DIFF
--- a/pkgs/bundle-image/builder.sh
+++ b/pkgs/bundle-image/builder.sh
@@ -69,6 +69,7 @@ echo "copying to image..."
 cptofs -p \
        -t ext4 \
        -i "$diskImage" \
+       --owner 11000 --group 11000 \
        "$root"/* / ||
     (echo >&2 "ERROR: cptofs failed. diskSize might be too small for closure."; exit 1)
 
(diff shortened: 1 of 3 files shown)


ryantm (approved):
Ran through the test plan and it looks good to me. Thank you!

ryantm:
Merged via https://github.com/replit/nixmodules/pull/131

PR #133, opened 2023-10-10
Template tests

airportyh:
Why
===

Want to run template tests as part of the deployment process.

What changed
============

1. add `test_templates` release channel to prodvana. It should be run after canary is deployed.
2. created test_templates.py script that calls template tester to run some tests
3. added test_templates.yaml kube manifest for running test_templates.py; nixmodules_version from semaphore is passed the script via `-m` option
4. Added a nixmodules namespace to k8s
5. copied git-ro-key to nixmodules namespace
6. added jobs-auth to nixmodules namespace

Test plan
=========

E2e test next time we build the disk.

Rollout
=======


manifests/application.pvn.yaml
```diff
@@ -29,6 +29,14 @@ application:
...
             - postDeployment:
                 checkDuration: 600s
+    - name: test_templates
+      preconditions:
+        - releaseChannelStable:
+          releaseChannel: canary
+      runtimes:
+        - name: config-raw
+          runtime: marine-cycle-160323--govalctl-config-raw
+          type: EXTENSION
```
pacharrin, line 39: if you want to run k8s we dont need the custom runtime
```suggestion
      runtimes:
          - runtime: marine-cycle-160323--goval
          containerOrchestration:
            k8s:
              namespace: ansible
```

manifests/application.pvn.yaml
```diff
@@ -49,7 +57,7 @@ application:
       preconditions:
         - manualApproval: {}
         - releaseChannelStable:
-            releaseChannel: canary
+            releaseChannel: template_tester
```
pacharrin, line 60: this should be the name of the release channel

```suggestion
            releaseChannel: test_templates
```

manifests/test_templates.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: batch/v1
```
pacharrin, line 1: have you run this manually with `kubectl create`? just to make sure this works without the added complexity of prodvana

manifests/test_templates.yaml
```diff
@@ -0,0 +1,63 @@
...
+kind: Job
+metadata:
+  name: test-templates
```
pacharrin, line 4:
```suggestion
  generateName: test-templates
```

prodvana will fail when they try apply this if we dont generate a different name for each run

scripts/test_templates.py
```diff
@@ -0,0 +1,81 @@
+#! /usr/bin/env nix-shell
+#! nix-shell -i python3 -p python311Packages.requests
+
+import requests
+from time import sleep
+import sys
+
+TEMPLATE_TESTER_URL = "https://templatetester.tobyho.repl.co"
+HEADERS = {
+  "Authorization": "BLARGH",
```
pacharrin, line 10: if this is a public repo we should create a semaphore secret for this

pacharrin (changes requested):

manifests/test_templates.yaml
airportyh: Yes. But probably want to use my own namespace instead of ansible.

scripts/test_templates.py
airportyh: Yup!

manifests/test_templates.yaml
```diff
@@ -0,0 +1,63 @@
...
+          volumeMounts:
+            - mountPath: /code
+              name: code
+          workingDir: /code/nixmodules
+      initContainers:
+        - args:
+            - "--ssh"
+            - "--repo=git@github.com:replit/nixmodules.git"
+            - "--root=/code"
+            - "--branch=th-tmpl-tests"
```
ryantm, line 34: Do we change this in a follow PR or now?

airportyh:
@pacharrin what would the flow look like if the tests failed?

ryantm (approved):

manifests/test_templates.yaml
airportyh: Thanks! Changed it to the rev from semaphore instead.

pacharrin (approved):

pacharrin:
> @pacharrin what would the flow look like if the tests failed?

it will let canary in the new version but wont try to push it to the other channels.


we could enable automatic rollbacks but that could also be a problem because during a rollback all RC are updated at once so canary wont be fully updated by the time the tests start

PR #137, opened 2023-10-12
use fenix for rust module

cdmistman:
Why
===

the rust module currently uses nixpkgs, but that comes with a few problems:
1. nixpkgs doesn't always get every version of the toolchain
  - for example, 1.72.1 has never been added to unstable, even though 1.73 is out now
  - 1.73 has been out for a week, but still isn't in unstable (which has a delay of ~36 hours iirc)
2. nixpkgs doesn't have every component in the toolchain
  - `rust-src` and `rust-std` aren't included, which means cross-compiling is impossible without overriding nixpkgs config attrs, which isn't possible
3. we don't update nixpkgs often, and it'd be nice if we could update the rust nixmodule independently of other nixmodules

What changed
============

- added input on [my fork of fenix](https://github.com/cdmistman/fenix) that has some patches so that the `version` attribute on various toolchain derivations are more 'accessible' (ie, it uses the tool's version number instead of the date the toolchain is built from)
- the module name no longer uses the rust version (when rust 2.0 drops we'll talk)
- changed rust nixmodule to use the fenix toolchain
- got rid of `cargo_run` script since we plan on adding support for project initializers
- got rid of `rustfmt` formatter since that's quite uncommon to invoke directly (when we add bazel support we'll talk)

Test plan
=========

with a blank repl:
- install the new nixmodule
- `cargo init` in the shell works
- open `src/main.rs`
- lsp status is green
- `println!` autocompletes
- run button works
- adding `eyre` via packages pane works

Rollout
=======

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,6 +2,8 @@
   description = "Nix expressions for defining Replit development environments";
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-23.05";
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+  inputs.fenix.url = "github:nix-community/fenix";
+  inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
   inputs.nixd.url = "github:nix-community/nixd";
   inputs.nixd.inputs.nixpkgs.follows = "nixpkgs";
   inputs.prybar.url = "github:replit/prybar";
@@ -13,7 +15,7 @@
   inputs.ztoc-rs.url = "github:replit/ztoc-rs";
   inputs.ztoc-rs.inputs.nixpkgs.follows = "nixpkgs";
 
-  outputs = { self, nixpkgs, nixpkgs-unstable, prybar, java-language-server, nixd, zig-overlay, ... }:
+  outputs = { self, nixpkgs, nixpkgs-unstable, prybar, java-language-server, nixd, zig-overlay, fenix, ... }:
     let
       mkPkgs = nixpkgs-spec: system: import nixpkgs-spec {
         inherit system;
@@ -23,6 +25,7 @@
           java-language-server.overlays.default
           nixd.overlays.default
           zig-overlay.overlays.default
+          fenix.overlays.default
         ];
         # replbox has an unfree license
         config.allowUnfreePredicate = pkg: builtins.elem (nixpkgs.lib.getName pkg) [
--- a/modules.json
+++ b/modules.json
@@ -554,5 +554,9 @@
   "zig-0.11:v1-20231014-15426ef": {
     "commit": "15426ef79793bf7c424eb40865d507eacfdd44e6",
     "path": "/nix/store/17ji0rxmrfz30sy2b60c65vjv5sh6ksz-replit-module-zig-0.11"
+  },
+  "rust-stable:v1-20231012-19c270f": {
+    "commit": "19c270f8c0c3147800ada35d9a0755fe10905344",
+    "path": "/nix/store/3hw6spngfg758gz8ygla1458phpdi6x5-replit-module-rust-stable"
   }
 }
--- a/pkgs/modules/rust/default.nix
+++ b/pkgs/modules/rust/default.nix
@@ -1,29 +1,16 @@
-{ pkgs-unstable, lib, ... }:
+{ pkgs, lib, ... }:
 let
-  pkgs = pkgs-unstable;
-  cargoRun = pkgs.writeScriptBin "cargo_run" ''
-    if [ ! -f "$HOME/$REPL_SLUG/Cargo.toml" ]; then
-      NAME=$(echo $REPL_SLUG | sed -r 's/([a-z0-9])([A-Z])/\1_\2/g'| tr '[:upper:]' '[:lower:]')
-      ${pkgs.cargo}/bin/cargo init --name=$NAME
-    fi
-
-    ${pkgs.cargo}/bin/cargo run
-  '';
-  rust-version = lib.versions.majorMinor pkgs.rustc.version;
+  inherit (pkgs.fenix.stable) toolchain;
 in
 {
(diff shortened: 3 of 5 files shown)


cdmistman:
everything works except rust-analyzer, i think because sysroot is broken. unfortunately, we don't show lsp logs :sweat_smile:

pkgs/modules/rust/default.nix
```diff
@@ -1,60 +1,42 @@
...
-    start = "${pkgs.cargo}/bin/cargo fmt";
-    stdin = false;
-  };
-
-  replit.dev.formatters.rustfmt = {
-    name = "rustfmt";
-    language = "rust";
-
-    start = "${pkgs.rustfmt}/bin/rustfmt $file";
+    start = "${toolchain}/bin/cargo fmt";
```
airportyh, line 39: The IDE wants to format just one file, so this should still work with a file formatter.

cdmistman: `rustfmt` alone doesn't respect `rustfmt.toml` configurations. can't we just ignore the `$file` env var?

ryantm, line 39: I don't think we should make this change, because right now I think we have the contract that our formatters are about formatting a single file. I think this formats the while project, which might be surprising and have an unexpected algorithmic complexity from the Workspace's perspective.

cdmistman:
@ryantm @airportyh since both of yall commented on the same thing:

- isn't the default formatter the lowest alphabetically? ie right now, there are 2 formatters for rust: `cargo-fmt` and `rustfmt`. since `cargo-fmt` is alphabetically "lower", aren't we already formatting the entire project with the rust nixmodule?
- `rustfmt` doesn't support automatically-reading `rustfmt.toml` (since it's implemented at the project level, and rustfmt doesn't know about rust projects)
- if a user intends for a project to be configured to be formatted a certain way, what's more surprising and/or damaging: that some files aren't formatted that way, or that files may unintentionally be formatted that way?

cdmistman:
cc @bradymadden97 toby mentioned you may have some more context on whether `cargo fmt` vs `rustfmt $file` will have serious impact on workspace performance

ryantm:
> * `rustfmt` doesn't support automatically-reading `rustfmt.toml`

It looks like it does if you pass --config-path to it:

```
        --config-path [Path for the configuration file]
                        Recursively searches the given path for the
                        rustfmt.toml config file. If not found reverts to the
                        input file path
```

cdmistman:
> It looks like it does if you pass --config-path to it:

that feels problematic 😅 prettier doesn't check `node_modules` but since rustfmt is unaware of project configurations, it'll check `/target` every time it's invoked

bradymadden97:
A few notes on how the workspace formatting works:
* The workspace expectation is that only the file you are looking at gets formatted. This is because we expect the formatter to return the formatted document as a string _to the workspace_, where the workspace handles locally mapping over any operations that happened _while_ formatting was being performed, and _then_ emitting the changes to the FS.
* If the formatter command directly modifies the FS, we will run the risk of race conditions (think: user A initiates "format", then continues typing while formatting occurring. If formatter writes results directly to FS - it will clobber those new local changes.)


I feel like this behavior also matches other IDEs - at least in vscode the commands are format document/selection, not anything across multiple files.

cdmistman:
after talking with @bradymadden97 offline as well it was noted that replit supports formatting via LSP, making the `formatter` configuration pointless. removing!

modules.json
```diff
@@ -554,5 +554,9 @@
   "zig-0.11:v1-20231014-15426ef": {
     "commit": "15426ef79793bf7c424eb40865d507eacfdd44e6",
     "path": "/nix/store/17ji0rxmrfz30sy2b60c65vjv5sh6ksz-replit-module-zig-0.11"
+  },
+  "rust-stable:v1-20231012-f561bea": {
+    "commit": "f561bea1c7b00ecc0be4040e3a0de993bdfc2834",
+    "path": "/nix/store/380kbyr6n0cmddspp1g7sz6nnixj4f57-replit-module-rust-stable"
```
ryantm, line 560: I think we need an upgrade map.

pkgs/modules/rust/default.nix
```diff
@@ -1,61 +1,35 @@
...
-    clang
-    rustc
-  ];
+  replit.dev.packages = with toolchain; [
+    toolchain
 
-  replit.dev.packages = with pkgs; [
-    rustfmt
-    rust-analyzer
+    pkgs.clang
```
ryantm, line 12: If we need this for the Rust, can we wrap rust instead so we don't leak all of clang into the environment?

ryantm: Okay if this is a followup since it already had this problem before.

flake.nix
```diff
@@ -2,6 +2,9 @@
   description = "Nix expressions for defining Replit development environments";
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-23.05";
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+  # TODO: switch to upstream once https://github.com/nix-community/fenix/pull/118 is merged
+  inputs.fenix.url = "github:cdmistman/fenix/for-nixmodules";
+  inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
```
ryantm, line 7: Since you decided to change to using the name "stable", do you still need your special patch?

ryantm (changes requested):

pkgs/modules/rust/default.nix
cdmistman: i don't think this makes sense. some rust tooling still relies on there being a c compiler in the environment eg during build scripts. sure, those will be available during the build scripts, but then it becomes invisible to the user which c compiler is being used.

ryantm: Okay, that sounds like a reasonable argument.

ryantm (approved):

PR #139, opened 2023-10-13
package json runner again

cdmistman:
Why
===

the last package.json runner script only works for bun and is frankly subpar. however the last time i introduced this package.json runner script everybody was sad because run button stopped working

What changed
============

brought back the new-and-improved package.json runner script but this time 👉😎👉 with no bugs

Test plan
=========

- WHEN TESTING IN CANARY MAKE SURE THE RIGHT DISK IS ATTACHED
- ALSO MAKE SURE YOU HAVE THE RIGHT MODULE INSTALLED
- try the matrix of setting the following: `scripts.replit-dev` in `package.json`, `scripts.dev` in `package.json`, `main` in `package.json`, and setting `entrypoint` in `.replit`

Rollout
=======

### DIFF
--- a/pkgs/modules/bun/default.nix
+++ b/pkgs/modules/bun/default.nix
@@ -6,23 +6,6 @@ let
   extensions = [ ".json" ".js" ".jsx" ".ts" ".tsx" ];
 
   community-version = lib.versions.majorMinor bun.version;
-
-  package-json-runner = pkgs.writeScript "bun-runner" ''
-    #!${pkgs.bash}/bin/bash
-    if start=$(${pkgs.jq}/bin/jq -r '.scripts.start? // empty' package.json 2>/dev/null); [[ -n $start ]]; then
-      echo "+ bun run start";
-      ${bun}/bin/bun run start;
-    elif module=$(${pkgs.jq}/bin/jq -r 'if .type == "module" then .module // empty else empty end' package.json 2>/dev/null); [[ -n $module ]]; then
-      echo "+ bun $module";
-      ${bun}/bin/bun $module;
-    elif main=$(${pkgs.jq}/bin/jq -r '.main // empty' package.json 2>/dev/null); [[ -n $main ]]; then
-      echo "+ bun $main";
-      ${bun}/bin/bun $main;
-    elif [[ -n $file ]]; then
-      echo "+ bun $file";
-      ${bun}/bin/bun $file;
-    fi
-  '';
 in
 
 {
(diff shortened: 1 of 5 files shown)


pkgs/modules/bun/default.nix
```diff
@@ -39,18 +26,14 @@ in
     bun
   ];
 
-  replit.dev.languageServers.typescript-language-server.extensions = extensions;
+  replit.languageServers.typescript-language-server.extensions = extensions;
```
ryantm, line 29: Is this an unintentional change? I think we don't want the language server config in a Deployment.

pkgs/modules/bun/default.nix
```diff
@@ -39,18 +26,14 @@ in
...
+  replit.runners."package.json" = {
     language = "javascript";
     inherit extensions;
-
-    start = "${package-json-runner}";
-    optionalFileParam = true;
   };
 
-  replit.dev.packagers.bun = {
+  replit.packagers.bun = {
```
ryantm, line 36: here too

ryantm (changes requested):

pkgs/modules/bun/default.nix
cdmistman: oh, it's unintentional yeah but i don't think we *ever* want the language server config in a deployment? why does lsp have to have `.dev`?

ryantm: Toby chose to make the options symmetrical between .dev and not .dev.

ryantm (approved):
I'm still a little concerned about how I couldn't really run the package-json-runner script to debug things because node freaked out about package.json files, but okay for now.

pkgs/upgrade-maps/default.nix
```diff
@@ -41,22 +41,22 @@ let
...
+      to = "bun-1.0:v8-20231013-913ea19";
+      auto = true;
+      changelog = ''# `package.json` runner
+        The runner for bun module has changed for increased flexibility and conformance to standards.
+
+        The new precedence is:
+        - the `replit-dev` script defined in `package.json`
+        - the `dev` script defined in `package.json`
+        - the `main` file defined in `package.json`
+        - the `entrypoint` defined in `.replit`
```
masad-frost, line 57: I would prefer if we flipped the order here since generally `main` is for node_modules imports and is compiled assets. `.replit`'s entrypoint is something we document and is a little more explicit.

cdmistman: done in [b900241](https://github.com/replit/nixmodules/pull/139/commits/b900241f49e1698496ddf03cccf9bf57b6b7d7c8)

PR #141, opened 2023-10-16
Python 3.11 yapf formatter fix

airportyh:
Why
===

Template tester found lsp formatter is broken in Python 3.11 template.

What changed
============

Use the version of yapf that matches the version of Python for pyright-extended.

Test plan
=========

Formatter should work for https://replit.com/@replit/Python-311

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -459,6 +459,10 @@
     "commit": "d4ad2e4453adf8dc310fe700f38d125361906606",
     "path": "/nix/store/wbxj48mmhp238zrdikdb86yz702s9llf-replit-module-python-3.11"
   },
+  "python-3.11:v7-20231016-35990c4": {
+    "commit": "35990c496af3e4a137a67c1c7a5ebec0849610b3",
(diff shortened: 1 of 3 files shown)


modules.json
```diff
@@ -483,6 +487,10 @@
     "commit": "7459cc2a5fd03bda902c2cd8d9d1fb387181b570",
     "path": "/nix/store/0jsq2yq7ds0szfhyrcsak99r79q7mf9c-replit-module-python-3.8"
   },
+  "python-3.8:v7-20231016-35990c4": {
+    "commit": "35990c496af3e4a137a67c1c7a5ebec0849610b3",
+    "path": "/nix/store/904k10vqg0avp4pg7fpzzma36p3jnw9v-replit-module-python-3.8"
+  },
```
ryantm, line 493: Why doesn't the python 3.10 module change?

airportyh: It must have already been using python310Packages.yapf which equates to nixpkgs.yapf.

PR #144, opened 2023-10-20
all the package managers

cdmistman:
Why
===

it doesn't make sense for only `npm` to be available in nodejs repls. users should have choice!

https://linear.app/replit/issue/DX-240/add-all-package-managers-to-path-for-js-nixmodules

couldn't get `bun` in there without adding pkgs-unstable to the derivation 🤷

What changed
============

added yarn and pnpm to PATH for nodejs modules

Test plan
=========

- `which yarn` in an empty repl with nodejs nixmodules installed works
- ditto with `which pnpm`

Rollout
=======

### DIFF
--- a/pkgs/modules/nodejs/default.nix
+++ b/pkgs/modules/nodejs/default.nix
@@ -1,16 +1,16 @@
 { nodejs }:
-{ pkgs, lib, ... }:
+{ pkgs, pkgs-unstable, lib, ... }:
 
 let
-
   community-version = lib.versions.major nodejs.version;
+
+  bun = pkgs-unstable.callPackage ../../bun { };
 
   nodepkgs = pkgs.nodePackages.override {
     inherit nodejs;
   };
 
   prettier = nodepkgs.prettier;
-
 in
 
 {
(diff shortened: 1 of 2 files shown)


ryantm:
> couldn't get bun in there without adding pkgs-unstable to the derivation 🤷

I think you might as well add it. If it is development only, it won't affect our disk building (which already has bun) and it shouldn't bloat Deployments.

pkgs/modules/nodejs/default.nix
```diff
@@ -23,9 +23,10 @@ in
   ];
 
   replit = {
-
     packages = [
       nodejs
+      nodepkgs.pnpm
+      nodepkgs.yarn
```
ryantm, line 29: let's add bun too!

ryantm (changes requested):

pkgs/modules/nodejs/default.nix
```diff
@@ -23,9 +23,10 @@ in
...
     ];
```
ryantm, line 31:
```suggestion
    ];
    dev.packages = [
      nodepkgs.pnpm
      nodepkgs.yarn
    ];
```

ryantm (changes requested):

pkgs/modules/nodejs/default.nix
ryantm: And can I think we should add bun too!

ryantm (commented):
Looks good. Now we need to update the lockfile and upgrade maps!

cdmistman:
oh i forgot to push the lockfile change. no need to update upgrade maps though since the only affected modules already have linear upgrades

ryantm (approved):

PR #146, opened 2023-10-23
ld library wrapper: introduce REPLIT_LD_LIBRARY_PATH and wrap nodejs

ryantm:
Why
===
* This wrapping solution continues to be a hack. We're only introducing this a temporary bandaid.
* We are noticing people have the same problem they have with Python, where they sometimes need to open a dynamic library dlopen in node. And using LD_LIBRARY_PATH is problematic, because it might interfere with other things running in the container.
* Choosing to use the name `REPLIT_LD_LIBRARY_PATH`, so that if we need to keep this hack around or expand it to further runtimes, we don't need to use a different name for each language. Also, this name makes it clear this is a Replit-specific thing.
* I plan to make env var work for Python in a follow-on.

What changed
===
* If REPLIT_LD_LIBRARY_PATH is set, set LD_LIBRARY_PATH to that. Otherwise, the wrapper does nothing.

Test plan
===

This should show the default LD_LIBRARY_PATH
```
node -e "console.log(process.env.LD_LIBRARY_PATH)"
```

This should print "a"
```
REPLIT_LD_LIBRARY_PATH=a node -e "console.log(process.env.LD_LIBRARY_PATH)"
```

Rollout
=======

### DIFF
--- a/pkgs/modules/nodejs/default.nix
+++ b/pkgs/modules/nodejs/default.nix
@@ -11,6 +11,25 @@ let
 
   prettier = nodepkgs.prettier;
 
+  ldLibraryPathWrapper = bin: pkgs.writeShellScriptBin (builtins.baseNameOf bin) ''
+    if [ -z ''${REPLIT_LD_LIBRARY_PATH+x} ]; then
+      :
+    else
+      export LD_LIBRARY_PATH=''${REPLIT_LD_LIBRARY_PATH}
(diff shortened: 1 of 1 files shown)


masad-frost (approved):

ryantm:
closing in favor of https://github.com/replit/nixmodules/pull/198

PR #148, opened 2023-10-24
update nixpkgs-unstable, add go-1.21 module

cdmistman:
Why
===

people want go 1.21 but it's not available on replit

What changed
============

- updated nixpkgs-unstable input
- added go 1.21 module

Test plan
=========

- sveltekit template tests pass
- bun template tests pass
- go template tests pass when set to go 1.21 module
  - note that this will require changing the template and running against the working release. 
  - also note that the template should not be released until enough conmen have attached to the new nixmodules disk build containing these commits

Rollout
=======

### DIFF
--- a/pkgs/modules/default.nix
+++ b/pkgs/modules/default.nix
@@ -35,7 +35,16 @@ let
     })
     (import ./nodejs-with-prybar)
 
-    (import ./go)
+    (import ./go {
+      inherit (pkgs) go gopls;
+    })
+    (import ./go {
+      go = pkgs-unstable.go_1_21;
+      gopls = pkgs-unstable.gopls.override {
+        buildGoModule = pkgs-unstable.buildGo121Module;
+      };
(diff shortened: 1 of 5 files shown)


pkgs/modules/default.nix
```diff
@@ -35,7 +35,14 @@ let
     })
     (import ./nodejs-with-prybar)
 
-    (import ./go)
+    (import ./go {
+      inherit (pkgs) go gopls;
+    })
+    (import ./go {
+      inherit (pkgs-unstable) gopls;
```
ryantm, line 42:
```suggestion
      gopls = pkgs-unstable.gopls.override { buildGoModule = pkgs.buildGo121Module; };
```
Let's make sure the gopls matches the same version as go.

ryantm (changes requested):

ryantm (approved):

PR #149, opened 2023-10-25
C/C++ to use fileParam

airportyh:
Why
===

These feel out of place that they hardcode the entrypoint instead of using a parameter. The reason I did it this way are irrelevant now.

What changed
============

Use the file parameter so that the entrypoint specified in .replit will be used.

Test plan
=========

1. install the C or C++ module
2. make a main.c or main.cpp file
3. set it as the entrypoint in a repl
4. reload the repl
5. click run and see it run


ryantm (approved):
Will this break templates that don't have the entrypoint specified as main.c? Still seems worth it for consistency and user flexibility.

airportyh:
> Will this break templates that don't have the entrypoint specified as main.c? Still seems worth it for consistency and user flexibility.

We don't have existing templates using this yet. I am about to make them.

pkgs/clang-compile/default.nix
```diff
@@ -36,14 +36,16 @@ pkgs.writeShellScriptBin "clang-compile" ''
...
     COMPILER="${clang}/bin/clang++"
+    COMPILER_NAME="clang++"
   else
     echo "Invalid LANG parameter: $LANG"
     exit 1
   fi
 
   rm -f ''$FILE.bin
-  set -o xtrace
+  echo $COMPILER_NAME $CFLAGS $SRCS -o "''$FILE.bin"
```
airportyh, line 49: Changed the print out to print short name of the binary instead of "scary" full nix path.

PR #152, opened 2023-10-26
[docker] Add the Docker module

lhchavez:
Why
===

We want to try seeing what would be needed to support running Docker containers in Replspace.

What changed
============

This change adds the necessary support to run containers in Replspace (with the exception of `replit-dockerd`, which will be provided elsewhere).

Test plan
=========

This was tested on one Repl. e2e test coming soon!

Rollout
=======

### DIFF
--- a/pkgs/modules/docker/default.nix
+++ b/pkgs/modules/docker/default.nix
@@ -1,0 +1,78 @@
+{ pkgs, pkgs-unstable, ... }:
+
+let
+
+  replit-runc = pkgs-unstable.buildGo121Module {
+    pname = "replit-runc";
+    version = "1.1.9+replit";
+
+    src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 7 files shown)


pkgs/modules/docker/default.nix
```diff
@@ -0,0 +1,68 @@
...
+    '';
+  };
+
+in
+
+{
+  id = "docker";
+  name = "Support for Docker containers";
+
+  replit.packages = [
```
ryantm, line 62: Do you need these all in deployments? If not, maybe replit.dev.packages for some of them.

ryantm (approved):

pkgs/modules/docker/default.nix
lhchavez: ah! i need _none_ in deployments!

PR #155, opened 2023-10-28
[rocker] Add the `docker` module to the `custom-bundle-squashfs`

lhchavez:
Why
===

We want to test this locally.

What changed
============

This change adds `docker` to the squashfs bundles.

Test plan
=========

```shell
make custom-nixmodules-disk
```

in conman, now i'm able to use

```toml
modules = ["docker:v1-20231026-3990bf0"]
```

in `.replit`.

Rollout
=======

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -93,7 +93,7 @@ rec {
     all-modules;
 
   bundle-squashfs = bundle-squashfs-fn {
-    moduleIds = [ "python-3.10" "nodejs-18" ];
+    moduleIds = [ "python-3.10" "nodejs-18" "docker" ];
     inherit upgrade-maps;
   };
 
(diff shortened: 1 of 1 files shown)


pkgs/default.nix
```diff
@@ -93,15 +93,15 @@ rec {
...
+    moduleIds = [ "python-3.10" "nodejs-18" "docker" ];
     inherit upgrade-maps;
   };
 
   custom-bundle-squashfs = bundle-squashfs-fn {
     # customize these IDs for dev. They can be like "python-3.10:v10-20230711-6807d41" or "python-3.10"
     # publish your feature branch first and make sure modules.json is current, then
     # in goval dir (next to nixmodules), run `make custom-nixmodules-disk` to use this disk in conman
-    moduleIds = [ "python-3.10" "nodejs-18" ];
+    moduleIds = [ "python-3.10" "nodejs-18" "docker" ];
```
airportyh, line 104: This is meant to be temporary so no need to check this in.

airportyh (approved):

pkgs/default.nix
```diff
@@ -93,15 +93,15 @@ rec {
     all-modules;
 
   bundle-squashfs = bundle-squashfs-fn {
-    moduleIds = [ "python-3.10" "nodejs-18" ];
+    moduleIds = [ "python-3.10" "nodejs-18" "docker" ];
```
airportyh, line 96: This will be used by all devs who don't build a custom nix modules disk.

pkgs/default.nix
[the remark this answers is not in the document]
lhchavez: also added a comment explaining not to check it in.

PR #158, opened 2023-10-30
Fix: renamed haskell-language-server to haskell-language-server-wrapper

airportyh:
Why
===

haskell-language-server is no longer in the bin after we upgraded somewhere along the way. But there is haskell-language-server-wrapper and haskell-language-server-9.0.2.

What changed
============

Referrence haskell-language-server-wrapper instead.

Test plan
=========

haskell lsp should work for `.hs` files


ryantm (approved):
built haskell-language-server locally and indeed it's gone.

PR #159, opened 2023-10-30
increase time limit; switched it to the job block

airportyh:
Why
===

https://github.com/replit/nixmodules/pull/157 didn't work

What changed
============

Moved it to the job block. This should work.

Test plan
=========

Try the build again.


ryantm (approved):
https://docs.semaphoreci.com/reference/pipeline-yaml-reference/#an-example-with-the-execution_time_limit-property-in-a-block looks like it should have worked before.

ryantm:
Outdated with new CI instance group

PR #160, opened 2023-10-30
maybe cache faster

ryantm:
Why
===
* Before we were having to evaluate the same flakes multiple times, this should limit the evaluation to just once per historical flake

What changed
===
* Add more debug output to build_disk_image.sh
* cache bundle-locked instead of each module separately
* Output the build of bundle-locked

test plan
===
* ci run

Rollout
=======


ryantm:
Nope, we went faster a different way.

PR #162, opened 2023-10-31
dont symlink python deps

cdmistman:
Why
===

https://linear.app/replit/issue/DX-297/nix-modules-breaks-django-within-python-repls

What changed
============

removed package symlinking for pip

Test plan
=========

- template tests work
- repro steps from linear ticket "with nixmodules" works

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -1,15 +1,13 @@
 { python, pypkgs }:
 { pkgs, lib, ... }:
 let
+  inherit (pypkgs) pip;
+
   pythonVersion = lib.versions.majorMinor python.version;
 
   pylibs-dir = ".pythonlibs";
 
   userbase = "$REPL_HOME/${pylibs-dir}";
-
-  pip = pkgs.callPackage ../../pip {
-    inherit pypkgs;
-  };
 
   pip-config = pkgs.writeTextFile {
     name = "pip.conf";
@@ -18,10 +16,6 @@ let
       user = yes
       disable-pip-version-check = yes
       index-url = https://package-proxy.replit.com/pypi/simple/
-
-      [install]
-      use-feature = content-addressable-pool
-      content-addressable-pool-symlink = yes
     '';
   };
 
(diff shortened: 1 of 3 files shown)


ryantm (commented):
Is this going to disable the Replit python package cache?

pkgs/modules/python/default.nix
```diff
@@ -18,10 +18,6 @@ let
       user = yes
       disable-pip-version-check = yes
       index-url = https://package-proxy.replit.com/pypi/simple/
```
airportyh, line 20: I am not ready to totally turn this off yet. Probably need to talk with the team. Can we disable for specific packages in https://github.com/replit/pip for now?

cdmistman:
> Is this going to disable the Replit python package cache?

i'm not sure! @airportyh do you know?

pkgs/modules/python/default.nix
cdmistman: i'd rather have python not be selectively broken for users. i think it's better to just broadly disable this and that way python Just Works on replit

airportyh:
> > Is this going to disable the Replit python package cache?
> 
> i'm not sure! @airportyh do you know?

Yes. The symlinking is what enables the cache.

ryantm (approved):

pkgs/modules/python/default.nix
airportyh: While we are at it, we can remove these lines:
```
      disable-pip-version-check = yes
      index-url = https://package-proxy.replit.com/pypi/simple/
```

lhchavez: if we have an allowlist for the packages we know it works, it wouldn't be "selectively broken": things that we don't know whether it works for will succeed, and the things we know it succeeds for would also succeed!

this might be a good middle ground in the short / medium term.

airportyh: Yeah, I agree. Especially packages known to be large.

PR #165, opened 2023-10-31
prune disk to just terminal and previous modules

ryantm:
Why
===
* CI was failing because it was running out of memory. This uses less memory, only about 13GB.
* Reduces the bundle lock closure size from 51GB to 41GB (19%)

What changed
===
* Moved mapping into its own file
* Added a passthru for upgrade-maps that lists the terminal and previous to terminal modules filtered out from the module lock file
* Use this list of module locks to generate the default module-lock filtering used by the main disk

Test plan
===
* CI passes
* Make sure template tests pass for all templates after deployed to canary

Rollout
=======

### DIFF
--- a/pkgs/upgrade-maps/default.nix
+++ b/pkgs/upgrade-maps/default.nix
@@ -2,101 +2,8 @@
 with pkgs.lib.attrsets;
 let
   fns = import ./fns.nix { inherit pkgs; };
-  mapping = {
-    # Examples:
-    #"bun-0.5:v1-20230522-49470df" = { to = "bun-0.5:v2-20230522-0f45db1"; auto = true; changelog = "A better lsp!"; };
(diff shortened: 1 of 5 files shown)


pkgs/upgrade-maps/default.nix
```diff
@@ -121,6 +28,24 @@ let
...
     destination = "/recommend-upgrade.json";
   };
+
+  modules = builtins.fromJSON (builtins.readFile ../../modules.json);
+
+  # A module is terminal if it doesn't have an upgrade mapping, or the
+  # upgrade mapping doesn't have auto = true.
+  isTerminal = mod: !((mapping.${mod} or false).auto or false);
+
+  moduleToTerminal =
```
airportyh, line 38: very slick!

masad-frost: 【﻿ｆｕｎｃｔｉｏｎａｌ　ｐｒｏｇｒａｍｍｉｎｇ】

ryantm:
Merged as part of #166

PR #166, opened 2023-10-31
[bundle-image] switch from ext4 to squashfs

ryantm:
This is built on top of #165

Why
===
* cptofs seemed to be getting stuck taking over 1 hr (before I stopped it)to build a 30 GB (before compression) disk image
* mksquashfs makes the disk much more quickly 11 minutes, and takes up less space (I think) at 11GB the code for mksquashfs is much simpler
* later we can most likely unify bundle-squashfs and bundle-image with some local dev tweaks 

What changed
===
* Rip out all the ext4 disk shenanigans
* Rip out lkl patch which isn't needed since we are not using cptofs
* use mksquashfs which is OP

Test plan
===
* We already know mksquashfs works in local dev and CI.
* After our back PR lands that accepts squashfs disks, we will try this in canary with the typical tests.

Rollout
=======


airportyh (approved):
:heart:

PR #167, opened 2023-11-01
[ci] use nix-community cache

ryantm:
Why
===
* Fenix, our Rust toolchain input, uses the nix-community cache
* We can avoid building Rust stuff and download it from there instead

What changed
===
* Added nix-community cache info to our cache setup script https://app.cachix.org/cache/nix-community

Test plan
===
* See less rust stuff being built in CI
* See CI times go down

Rollout
=======


cdmistman:
maybe we should put all the substituters in the flake attribute `nixConfig`?

cdmistman (approved):
would be nice to use flake-level `nixConfig` attr for these substituters but that can be done separately

airportyh (approved):

PR #168, opened 2023-11-01
prune /etc files, refactor/simplify bundle

ryantm:
Why
===
* We started pruning the /nix/store paths of some modules, but we didn't remove them from modules.json and active-modules.json

What changed
===
* unify bundle-squashfs and bundle-image (parameterized by a diskname because local expects disk.sqsh and Google needs disk.raw for production) We could still fix our local version to support disk.raw in a followup
* Move the /etc/nixmodules definitions into the bundle-lock derivation, where they can be synced with the moduleLocks
* Modify bundle-image builder to copy the new /etc/ directory instead of copying files from a bunch of different derivations.
* Rearrange pkgs/default.nix a bit

Test plan
===
* Mounted disk locally and saw /etc/nixmodules had pruned files

Rollout
=======


airportyh (approved):
built `bundle-locked` and get `etc/nixmodules` within it. Files look good.

PR #170, opened 2023-11-03
keep all package managers in prod env

cdmistman:
Why
===

see https://replit.slack.com/archives/C03KS2B221W/p1699033526063659

users should be able to eg `yarn run prod` in their deployments

What changed
============

- took all package managers out of `dev.packages` and into `packages` directly
- this was a bit more involved for rust due to rust tooling making assumptions about the colocation of library/sources and tools

Test plan
=========

only rust needs to be tested, since the other changes were just removing restrictions on the PATH

since `rust-std`, `rustc`, and `cargo` are the only components necessary to get a basic hello world out, and the other components are for very, *very* specific scenarios (the only notable exception being cross-compiling, which we don't support [yet]), i think it's fine to restrict rust nixmodule to only testing:

- in the workspace:
- run button works
- lsp works
- format works
- `cargo fmt` works
- `clippy` works
- `cargo doc` works

Rollout
=======

### DIFF
--- a/pkgs/modules/rust/default.nix
+++ b/pkgs/modules/rust/default.nix
@@ -1,16 +1,26 @@
-{ pkgs, lib, ... }:
+{ pkgs, ... }:
 let
-  inherit (pkgs.fenix.stable) toolchain;
+  inherit (pkgs.fenix) stable;
 in
 {
   id = "rust-stable";
   name = "Rust Tools";
 
-  replit.dev.packages = with toolchain; [
-    toolchain
+  replit.packages = [
+    (stable.withComponents [
+      "cargo"
+      "llvm-tools"
+      "rust-src"
+      "rust-std"
+      "rustc"
+    ])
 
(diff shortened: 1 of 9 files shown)


lhchavez (approved):

pkgs/modules/rust/default.nix
```diff
@@ -1,18 +1,28 @@
-{ pkgs, lib, ... }:
+{ pkgs, ... }:
 let
-  inherit (pkgs.fenix.stable) toolchain;
+  toolchain = pkgs.fenix.stable;
```
ryantm, line 3: it might be confusing to name this toolchain when we later call toolchain.toolchain

Can we call this rust_stable or something? Not blocking since you know this ecosystem better than me.

ryantm (approved):
Aside from that rust comment, looks good. I grepped for other dev.packages and looked at them all and they seem good.

PR #171, opened 2023-11-06
only use replit pip when __REPLIT_PIP_CACHE_ENABLE is set

cdmistman:
Why
===

today during sync we decided that disabling the pip caching stuff should be flagged.

What changed
============

- always invoke our fork of pip
- if the env var `__REPLIT_PIP_CACHE_ENABLE` is set, then use a `pip.conf` file that enables our cacache symlinking behavior
- if the env var `__REPLIT_PIP_CACHE_ENABLE` isn't set, then use a `pip.conf` file that doesn't enable our cacache symlinking behavior

Test plan
=========

1. in a django repl, use this nixmodule and make sure things are installed etc etc
1. in `.replit` add `env.__REPLIT_PIP_ENABLE_CACHE_WRAPPER = "1"`
1. `poetry install`
1. `python3 manage.py migrate`
1. `python3 manage.py createsuperuser`
1. error!
1. remove the env var from step 2
1. repeat steps 3-5
1. no error!

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -359,6 +359,10 @@
     "commit": "d0fd0d762ab9adb644819e1974387ff37564b456",
     "path": "/nix/store/jvnsc9r7fzdxslwzsfzjmz34gagf1lzw-replit-module-python-3.10"
   },
+  "python-3.10:v29-20231107-a9f4961": {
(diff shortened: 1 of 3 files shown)


airportyh (approved):
lgtm

ryantm (approved):
Nice, I like how you extracted it into its own file.

PR #172, opened 2023-11-08
[docker] A few updates

lhchavez:
Why
===

There are a few things that don't quite work today for `docker-compose`.

What changed
============

This change:

* Cherry-picks a `buildkitd` panic that was fixed in
  https://github.com/moby/buildkit/commit/9a8bd9fde84da8190b9bb1cbdbe72ec843780e64
* The `containerd.toml` accidentally skipped the uid setting for the
  socket.
* Adds `docker-compose` to the bundle.
* Makes `replit-runc` resolve the `roci` binary: it's useful to be able
  to override it in Replspace.

Test plan
=========

Build custom modules in conman, see it work more.

Rollout
=======


ryantm (approved):
Looks good

PR #175, opened 2023-11-18
[Prodvana] Allow different runtime for test-templates

pacharrin:
Why
===

The current manifest was invalid, if you need one channel to not use the runtime extension, you cant define a global runtimeextension, you need to define it for every release channel

What changed
============

set service config for each release channel

Test plan
=========

`pvnctl config validate manifests` 

Rollout
=======


ryantm (approved):

airportyh (approved):
Thanks!

PR #177, opened 2023-11-20
[cd] Make sure that govalctl can deploy nixmodules configs

lhchavez:
Why
===

We had a bug where the manifest never overrode the default service name of `pid1`, so we were trying to push the pid1 configuration with nixmodules contents. That was not going to work!

What changed
============

This change correctly overrides the serviec name so that it's `nixmodules`.

Test plan
=========

:eyes: on Prodvana.

Rollout
=======


airportyh (approved):
Thanks!

ryantm (approved):

PR #185, opened 2023-11-28
[docker] Prevent any net sysctls

lhchavez:
Why
===

It turns out that there are a few differences between dev VMs and CI
where Docker tries to adjust some sysctls.

What changed
============

This change prevents any net-related sysctls from happening.

Test plan
=========

```
failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: sysctl "net.ipv4.ip_unprivileged_port_start" not allowed in host network namespace: unknown
```

no longer happens in CI.

Rollout
=======


ryantm (approved):
looks fine

masad-frost (approved):
:ghost:

PR #187, opened 2023-11-29
make things stable again

cdmistman:
Why
===

i made an untested assumption that `bootstrapped-pip` (which was removed from nixpkgs) works close enough to how `pip` works and that resulted in `pip` not working with nixmodules when testing a deploy :(

What changed
============

reverted #184 and re-locked modules from #186 

Test plan
=========

- python works again
- ci passes
- template tests pass
- luis's docker stuff works

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -722,5 +722,9 @@
   "docker:v6-20231128-8cef170": {
     "commit": "8cef17041b553bad202c306e5d9b726ddd007544",
     "path": "/nix/store/mb6v56in36xplrs8783q9923cdimdgvd-replit-module-docker"
+  },
+  "docker:v7-20231129-8a00f4c": {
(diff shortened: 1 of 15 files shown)


modules.json
```diff
@@ -35,10 +35,6 @@
     "commit": "f6dc97c19cbc64d2a271be3ba829b87fd5237328",
     "path": "/nix/store/62vlf5i47d5d60181w31n6zf1xfy2s0y-replit-module-c-clang14"
   },
-  "c-clang14:v4-20231128-1e033d9": {
```
lhchavez, line 38: spooktastic

lhchavez (approved):

modules.json
cdmistman: 👻

PR #189, opened 2023-11-30
Added npx wrapper that would ensure .config/npm/node_global/lib exists

airportyh:
Why
===

If you start with an empty repl and add the nodejs module, then run `npx create-react-app`, npx would try to install executables globally in `$REPL_HOME/.config/npm/node_global/lib` and would fail if it didn't exist. This ensures that directory exists.

What changed
============

Created a wrapper for npx that would ensure that directory.

Test plan
=========

1. create blank repl
2. add nodejs module
3. `npx create-react-app` and see it work


ryantm (approved):
Maybe you could use XDG_CONFIG_HOME instead of REPL_HOME?

PR #190, opened 2023-11-30
fix haskell lsp

cdmistman:
Why
===

currently the haskell module uses `haskell-language-server-wrapper` which has an infinite spinner in the lsp indicator. i think this is because the wrapper is doing something weird when trying to [pick the right GHC from PATH](https://github.com/NixOS/nixpkgs/blob/0b35753f0128c4a34d6712e60b0a23f893665f31/doc/languages-frameworks/haskell.section.md?plain=1#L769-L771)?

What changed
============

don't use `haskell-language-server-wrapper`, just use `haskell-language-server`

Test plan
=========

1. open haskell repl
2. open `Main.hs`
3. lsp spinner stops spinning eventually

Rollout
=======

### DIFF
--- a/pkgs/modules/haskell/default.nix
+++ b/pkgs/modules/haskell/default.nix
@@ -32,6 +32,6 @@ in
     name = "Haskell Language Server";
     language = "haskell";
 
-    start = "${pkgs.haskell-language-server}/bin/haskell-language-server-wrapper --lsp";
+    start = "${pkgs.haskellPackages.haskell-language-server}/bin/haskell-language-server --lsp";
   };
 }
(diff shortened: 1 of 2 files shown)


pkgs/modules/haskell/default.nix
```diff
@@ -32,6 +32,6 @@ in
     name = "Haskell Language Server";
     language = "haskell";
 
-    start = "${pkgs.haskell-language-server}/bin/haskell-language-server-wrapper --lsp";
```
ryantm, line 35: Have you tested this? It doesn't look like that binary you want to call exists, but maybe it depends on the version?

```
$ ls /nix/store/l24fg4ymhrzwxwlcxh5jmq92sf0g8z2z-haskell-language-server-2.4.0.0/bin/
haskell-language-server-9.4.8  haskell-language-server-wrapper
```

ryantm (changes requested):

ryantm (approved):
Thanks!

PR #194, opened 2023-12-01
make things unstable again

cdmistman:
Why
===

we want to move towards always having the latest and greatest instead of pinning to a convenient nixpkgs channel. this is a more moderate attempt at #184 that doesn't attempt to remove all of `nixos-23.05` but instead makes `nixpkgs-unstable` the default pkgs channel to us in modules.

as such, we'll be able to maintain `java` and (more importantly) all python nixmodules until we find a way to build them with unstable

What changed
============

- updated `nixpkgs-unstable` input
- set `pkgs` to `nixpkgs-unstable` for all nixmodules
- set `pkgs-23_05` to `nixos-23.05`

Test plan
=========

- node/python tests pass

Rollout
=======

### DIFF
--- a/pkgs/modules/typescript-language-server/default.nix
+++ b/pkgs/modules/typescript-language-server/default.nix
@@ -6,8 +6,7 @@ let
     nativeBuildInputs = [ pkgs.makeWrapper ];
     postInstall = ''
       wrapProgram "$out/bin/typescript-language-server" \
-        --suffix PATH : ${pkgs.lib.makeBinPath [ nodepkgs.typescript ]} \
-        --add-flags "--tsserver-path ${nodepkgs.typescript}/lib/node_modules/typescript/lib/"
+        --suffix PATH : ${pkgs.lib.makeBinPath [ nodepkgs.typescript ]}
     '';
   };
 in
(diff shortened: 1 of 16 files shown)


ryantm (approved):
thanks, looks good

pkgs/modules/typescript-language-server/default.nix
```diff
@@ -6,8 +6,7 @@ let
     nativeBuildInputs = [ pkgs.makeWrapper ];
     postInstall = ''
       wrapProgram "$out/bin/typescript-language-server" \
-        --suffix PATH : ${pkgs.lib.makeBinPath [ nodepkgs.typescript ]} \
-        --add-flags "--tsserver-path ${nodepkgs.typescript}/lib/node_modules/typescript/lib/"
+        --suffix PATH : ${pkgs.lib.makeBinPath [ nodepkgs.typescript ]}
```
ryantm, line 9: This doesn't work:
```
[languages]

[languages.javascript]
pattern = "**/{*.js,*.jsx,*.ts,*.tsx,*.json}"

[languages.javascript.languageServer]
start = "/nix/store/vmd368fmaklyyj232fvnrawqffd4zwq4-nodejs-20.3.1/bin/node /nix/store/gx32fpn4758z67wa4hqc4wcd7mqrvl9z-typescript-language-server-3.3.2/bin/.typescript-language-server-wrapped --stdio"
```
Is this a necessary change in this PR? I wonder if we can do this separately for easy of rollbacks.

ryantm (changes requested):

ryantm (approved):
Looks good

PR #198, opened 2023-12-06
use REPLIT_LD_LIBRARY_PATH for node

cdmistman:
Why
===

we're fully embracing `REPLIT_LD_LIBRARY_PATH`, so we should support it for our node nixmodules

What changed
============

when exec'ing node, prepend the value of `REPLIT_LD_LIBRARY_PATH` to `LD_LIBRARY_PATH`

Test plan
=========

run button for [this repl](https://replit.com/@coltonatreplit/NeighboringHonestRegression) works

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -243,6 +243,10 @@
     "commit": "9a12353905c85260e57f6533f3f38b8e304a3087",
     "path": "/nix/store/lj77k1jq6igdm7iz5fav95f7r9dvb53k-replit-module-nodejs-18"
   },
+  "nodejs-18:v18-20231206-d8b5d0b": {
+    "commit": "d8b5d0b9427bb53d86e0264499452e734e0f4555",
+    "path": "/nix/store/6khbf7ciw8x1ffg4arad3bfw0pbzapdg-replit-module-nodejs-18"
+  },
   "nodejs-19:v1-20230525-c48c43c": {
     "commit": "c48c43c6c698223ed3ce2abc5a2d708735a77d5b",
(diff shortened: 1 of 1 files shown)


pkgs/modules/nodejs/default.nix
```diff
@@ -2,6 +2,26 @@
...
+    name = "nodejs-wrapped";
+    buildInputs = [ pkgs.makeWrapper ];
+    buildCommand = ''
+      mkdir -p $out/bin
+      for bin in ${nodejs}/bin/*; do
+        local binName=$(basename $bin)
+        cat >$out/bin/$binName <<-EOF
+        if [ -z "$REPLIT_LD_LIBRARY_PATH" ]; then
+          export LD_LIBRARY_PATH="\$REPLIT_LD_LIBRARY_PATH:\$LD_LIBRARY_PATH"
+        fi
```
ryantm, line 15: I think the logic here is reversed. You're saying if REPLIT_LD_LIBRARY_PATH has zero length, add it to the path.

```suggestion
    if [ -z ''${REPLIT_LD_LIBRARY_PATH+x} ]; then
      :
    else
      export LD_LIBRARY_PATH="''${REPLIT_LD_LIBRARY_PATH}:''${LD_LIBRARY_PATH}"
    fi
```

ryantm (changes requested):

pkgs/modules/nodejs/default.nix
cdmistman: done!

PR #202, opened 2023-12-11
initial replit module

cdmistman:
Why
===

we want LSP support for `.replit` and we want it now!

note: unfortunately taplo uses a lot of memory :/ 
image

it's my intention to keep this feature-gated for staff only for now. i have a suspicion i can get this memory usage lower by making a custom binary that consumes only taplo's lsp crate and only loads the configuration for `.replit` into memory

another optimization that will come: before rolling this nixmodule out to GA, I plan on implementing lazy-loaded LSP servers. ideally this lsp config is only used when a user opens a `.replit` file

What changed
============

- added JSON schema for `.replit` file
- added novel `replit` nixmodule
- added LSP configuration for `replit` module to use `taplo` from [my PR](https://github.com/tamasfe/taplo/pull/510) for `.replit` files

Test plan
=========

- add `replit` nixmodule to a repl
- open `.replit` file
- lsp works

Rollout
=======

### DIFF
--- a/pkgs/modules/replit/dotreplit-schema.json
+++ b/pkgs/modules/replit/dotreplit-schema.json
@@ -1,0 +1,661 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$ref": "#/definitions/DotReplit",
+  "definitions": {
+    "DotReplit": {
+      "title": ".replit file",
+      "description": "Configuration for your Replit Repl.",
+      "type": "object",
+      "properties": {
+        "onBoot": {
+          "$ref": "#/definitions/Command"
+        },
+        "run": {
+          "$ref": "#/definitions/Command"
+        },
+        "compile": {
+          "$ref": "#/definitions/Command"
+        },
+        "language": {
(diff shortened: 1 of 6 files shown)


pkgs/modules/replit/dotreplit-schema.json
```diff
@@ -0,0 +1,661 @@
...
+      }
+    },
+
+    "NixConfig": {
+      "title": "Nix Configuration",
+      "description": "Configure Nix on Replit.",
+      "type": "object",
+      "properties": {
+        "channel": {
+          "enum": ["unstable", "stable-21_11", "stable-22_05", "stable-22_11", "stable-23_05"]
```
ryantm, line 502: darn, another place to update the channels.

cdmistman: yeah - even though it doesn't _have_ to be specified as an enum, it'd be nice for users to be able to see exactly what options they have

ryantm (approved):
Let's go!

PR #203, opened 2023-12-11
Poetry perf enhancement

airportyh:
Why
===

Poetry install in free repls are slow due to limited resources. This brings in https://github.com/replit/poetry/pull/27 to improve it by between 30%-50%.

What changed
============

1. updated poetry bundle urls to new ones
2. if repl only has 0.5 vCPUs turn off parallel workers for poetry

There's some pending clean up we can do such as just use the stock pip instead of custom. The next time we enable pip cache we'll need to modify the installer package.

Test plan
=========

1. re-test performance and see numbers are similar to https://docs.google.com/spreadsheets/d/1azEwc2Cd--b6YQl_h6xM33l1vQ8uf_YAhRwS6DAPy8k/edit#gid=0
2. do some regression tests on a list of common pkgs and make sure we didn't break anything obvious


pkgs/poetry/poetry-in-venv.nix
```diff
@@ -1,23 +1,32 @@
...
+    inherit version;
 
-  src = builtins.fetchTarball {
-    inherit url sha256;
-  };
-
-  buildInputs = [ pypkgs.pip ];
+    src = builtins.fetchTarball {
+      inherit url sha256;
+    };
```
airportyh, line 9: Removed pip as a buildInput here. Didn't need it anyway and it was conflicting with something.

pkgs/poetry/poetry-in-venv.nix
```diff
@@ -1,23 +1,32 @@
...
+                                # https://github.com/replit/poetry/blob/replit-1.1/poetry/utils/env.py#L885
+                                # invoking the workaround so that poetry
+                                # does not use its own venv for the project
+                                # env
+      $out/env/bin/pip install poetry --find-links ./ --no-index
+      ln -s $out/env/bin/poetry $out/bin/poetry
+    '';
+  };
+in
+pkgs.writeShellScriptBin "poetry" ''
```
airportyh, line 24: Yet another wrapper.

airportyh: Wait... this could hurt poetry startup performance.

airportyh: But not sure what else I can do.

pkgs/poetry/poetry-in-venv.nix
cdmistman: oooooh nice i think this unblocks the python modules being moved to nixpkgs-unstable

pkgs/poetry/poetry-in-venv.nix
cdmistman: image

as long as doing this yields >40ms, i think it's fine :)

cdmistman (approved):
lgtm 💨

PR #208, opened 2023-12-15
se new poetry version; modern installer + curl for poetry

airportyh:
Why
===

What to re-land https://github.com/replit/nixmodules/pull/203.

What changed
============

Same as previous, only different versions of poetry bundles.


cdmistman (approved):
🤞

PR #209, opened 2023-12-15
[Prettier] Support local prettier

bradymadden97:
Why
===

If a user has prettier installed locally (npm g or even local npm), we want to use that instead of the nixmodule prettier, incase they are different versions. (Maybe at some point we'll let them pick in the UI, but this is fine for now, we can assume this is correct)

So instead of just running prettier with the nixmodule path we'll run a script to find the right path of prettier, and then run that one.

I suppose this has the downside of making every formatter run a bit slower, as it has to run `which` every time. If you have a better idea than this, I'm all ears. I didn't want to compute it ahead of time, as it seemed like then it would be a challenge to re-calculate whenever the user's environment changed, like if they later installed or uninstalled prettier somewhere.

_Describe what prompted you to make this change, link relevant resources_

What changed
============

Run a script to pick the right prettier binary to run

Test plan
=========

I followed the steps in `docs/nixmodules-disks.md`
* Built a local nix module with my changes
* Ran goval locally with the custom nix module
* Loaded up replit web with a repl using the node js nix module

Then I tested the following sequence:

* Only the prettier nix package installed, which is v3.1. I set up a formatting change that had a behavior change in v3.0.2:
https://github.com/prettier/prettier/blob/main/CHANGELOG.md#302

* it formatted w/ prettier 3.1
* then I npm i -g prettier @ 2.8.4
* then formatted again, and saw it formatted the old way, indicating it ran version 2.8.4

https://github.com/replit/nixmodules/assets/16962017/8f11f71a-612c-4a6a-9ddc-c9068dc4a314

Rollout
=======

### DIFF
--- a/pkgs/modules/nodejs/default.nix
+++ b/pkgs/modules/nodejs/default.nix
@@ -31,8 +31,6 @@ let
     inherit nodejs;
   };
 
-  prettier = nodepkgs.prettier;
-
   npx-wrapper = pkgs.writeShellScriptBin "npx" ''
     mkdir -p ''${XDG_CONFIG_HOME}/npm/node_global/lib
     ${nodejs-wrapped}/bin/npx "$@"
@@ -57,7 +55,7 @@ in
     ];
 
     dev.packages = [
-      prettier
+      nodepkgs.prettier
     ];
 
     dev.languageServers.typescript-language-server.extensions = [ ".js" ".jsx" ".ts" ".tsx" ".json" ".mjs" ".cjs" ".es6" ];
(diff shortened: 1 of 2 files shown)


pkgs/modules/nodejs/default.nix
```diff
@@ -118,7 +121,7 @@ in
       language = "javascript";
       extensions = [ ".js" ".jsx" ".ts" ".tsx" ".json" ".html" ];
       start = {
-        args = [ "${prettier}/bin/prettier" "--stdin-filepath" "$file" ];
+        args = [ "${prettierScript}/bin/prettier" "--stdin-filepath" "$file" ];
```
cdmistman, line 124: let's just change this to this
```suggestion
        args = [ "prettier" "--stdin-filepath" "$file" ];
```

cdmistman (approved):

PR #214, opened 2023-12-16
Repair modules.json lock file

airportyh:
Why
===

We published broken modules (in particular the ones depending on poetry changes) to modules.json. This change removes those modules and runs lock_modules.py again to put in the correct versions of those modules.

What changed
============

1. removed broken modules
2. python scripts/lock_modules.py

Test plan
=========

1. disk should build; ocis should build

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,3 +204,4 @@ python scripts/lock_modules.py
 git commit -am "updated X in modules.json"
 git push
 ```
+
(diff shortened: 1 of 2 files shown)


ryantm (approved):
Thanks

blast-hardcheese (approved):

PR #215, opened 2023-12-19
use typescript-language-server 4.2.0

cdmistman:
Why
===

ideally typescript-language-server would use `typescript` from the current project's `node_modules` but we haven't been doing that, since it's not guaranteed all repls will have that dep. so instead i submitted a path upstream that allows us to do that by setting `tsserver.fallbackPath` in initializationOptions but that's not in nixpkgs yet (it comes with 4.2.0, nixpkgs currently only has 4.1.2)

What changed
============

- override the `typescript-language-server` drv with the 4.2.0 bundle
- use `tsserver.fallbackPath` in initialization options instead of `tsserver.path`

Test plan
=========

- js completions work in a js-only project without the `typescript` dependency anywhere
- ^ but with `typescript` devDependency (`kill 1` after adding the dep to restart the lsp with the right tsserver)
- verify that the node_modules tsserver is actually being used by using a tsserver older than 5.1 (probably just do `"typescript": "^4.0.0"` and observing this code:
```ts
function foo(): undefined {}
```

if there's no error, then you're using nixpkgs-defined typescript. if there is an error (and you're correctly using the outdated version of `typescript`) then you're using the node_modules typescript!

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -263,6 +263,10 @@
     "commit": "fb57c712f0b9b0c1dd1ce912fc0b8b17855137fe",
     "path": "/nix/store/m2j90520zk4a76zwdpsq4c6fab4zaray-replit-module-nodejs-18"
   },
+  "nodejs-18:v23-20231219-faac932": {
+    "commit": "faac932188739c982a79d5108abeee749b2f445e",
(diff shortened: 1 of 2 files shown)


cdmistman:
happy holidays @airportyh

blast-hardcheese (approved):
Makes sense. Good stuff.

PR #217, opened 2023-12-19
deno module

cdmistman:
supercedes #207 

Why
===

denosaur says hello, world!

What changed
============

added new deno module (which is a lot simpler than #207 which was using a wrapper script which is kinda unnecessary now because deno now has support for `tasks.json` file)

Test plan
=========

- lsp works
- run button works

Rollout
=======

### DIFF
--- a/pkgs/modules/deno/default.nix
+++ b/pkgs/modules/deno/default.nix
@@ -1,0 +1,32 @@
+{ pkgs, lib, ... }:
+
+let
+  inherit (pkgs) deno;
+  version = lib.versions.major deno.version;
+
+  extensions = [ ".json" ".jsonc" ".js" ".jsx" ".ts" ".tsx" ];
+in
+
+{
+  id = "deno-${version}";
+  name = "Deno Tools";
+
+  replit.packages = [
+    deno
+  ];
+
+  replit.runners.deno-script-runner = {
+    name = "deno";
+    language = "javascript";
+    inherit extensions;
+    fileParam = true;
+    start = "${deno}/bin/deno run --allow-all $file";
+  };
+
+  replit.dev.languageServers.deno = {
+    name = "deno";
+    language = "javascript";
+    inherit extensions;
+    start = "${deno}/bin/deno lsp --quiet";
+  };
+}
--- a/modules.json
+++ b/modules.json
@@ -1026,5 +1026,9 @@
   "bash:v1-20231215-e6d471c": {
     "commit": "e6d471c9fd81b6ca469e1c42692953146c7cfb20",
     "path": "/nix/store/0j5kz53myr8kycfc4w1vkfccz4m5rsyv-replit-module-bash"
+  },
+  "deno-1:v1-20231215-5d427a8": {
+    "commit": "5d427a83eae4fc1be41c3842b294c98f41271462",
+    "path": "/nix/store/z02na0yx2xlnrkaaf1k515ddnib8hd2h-replit-module-deno-1"
   }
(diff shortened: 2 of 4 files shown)


masad-frost (approved):

pkgs/modules/deno/default.nix
```diff
@@ -0,0 +1,32 @@
...
+  replit.packages = [
+    deno
+  ];
+
+  replit.runners.deno-script-runner = {
+    name = "deno";
+    language = "javascript";
+    inherit extensions;
+    fileParam = true;
+    start = "${deno}/bin/deno run $file";
```
masad-frost, line 23: I wonder if we should allow all access by default

masad-frost (commented):
Maybe we should allow everything by default? :sauropod:

pkgs/modules/deno/default.nix
cdmistman: so i decided not to, because that's contra the deno philosophy. if users want to have some perms enabled, then they should do so via `tasks` in their `deno.json`

pkgs/modules/deno/default.nix
```diff
@@ -0,0 +1,32 @@
...
+  id = "deno-${version}";
+  name = "Deno Tools";
+
+  replit.packages = [
+    deno
+  ];
+
+  replit.runners.deno-script-runner = {
+    name = "deno";
+    language = "javascript";
```
ryantm, line 20: Why is it JavaScript here and below?

cdmistman: why wouldn't it be? we use javascript for everything that works with both js and ts...

cdmistman: this `language` field is completely arbitrary and i think just a holdout from polygott. the only reason i set it here is because i'm not confident things won't work without this field

pkgs/modules/deno/default.nix
masad-frost: Mmm, maybe. For Replit specifically the environment is sandboxed that this might be overkill, it feels like we will frustrate people instead of delighting them. Couldn't people also use tasks to run sandboxed tasks?

cdmistman: that's a great point! i'll go ahead and make the default `--allow-all`

pkgs/modules/deno/default.nix
blast-hardcheese: The "language: javascript" stuff may be better served by a set, where the set of all attributes of an environment can be tested if we know we need one thing or another, matching the most specific subset that we can find.

`{javascript, deno}` being distinct from `{javascript, typescript}` but still having some commonalities depending on what we need.

This is also a problem in upm, which is largely where I'm coming from here.

blast-hardcheese (approved):
looks alright

pkgs/modules/deno/default.nix
cdmistman: force-pushed with this change

pkgs/modules/deno/default.nix
cdmistman: @blast-hardcheese the worst part is, this instance of "javascript" shouldn't actually be anywhere in any of our codebases. it could be `coltonrulesz\\foobar` and wouldn't make a difference. the only reason it exists is for mapping glob patterns to language servers when they're being configured via .replit. it's an antiquated pigeonhole that serves an entirely different purpose with no beneficial heuristics we can't attach via other means

PR #221, opened 2024-01-02
adds the prerelease RG cluster to prodvana

ksh-chy:
Why
===

https://linear.app/replit/issue/INC-2264/inc-1238-ai-follow-up-on-why-prerelease-has-a-very-old-nixmodules-disk

What changed
============

Adds the prerelease cluster to the Prodvana manifests, updates the dependencies from `picard` ->  `prerelease` from the rollout. 

Test plan
=========

`prerelease` is in the [config bucket already](https://console.cloud.google.com/storage/browser/replit-config/nixmodules/prerelease), it just hasn't been updated since November at the latest. 

Rollout
=======

We will just need to monitor the rollout plan and ensure the config bucket + prerelease blob look good.

### DIFF
--- a/manifests/application.pvn.yaml
+++ b/manifests/application.pvn.yaml
@@ -107,7 +107,6 @@ application:
         - name: config-raw
           runtime: marine-cycle-160323--govalctl-config-raw
           type: EXTENSION
-# regional goval
     - name: staging
       constants:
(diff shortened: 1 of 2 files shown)


ryantm (approved):
Thanks

PR #222, opened 2024-01-02
shellcheck all the things

ryantm:
Why
===
* We had a quoting issue that shellcheck would have caught
* Linting is a good practice, especialy with all the footguns in Shell

What changed
===
* convert all writeShellScriptBin uses to writeShellApplication
* add CI check to disallow writeShellScriptBin

Test plan
===
* If CI passes and the disk builds, this should be okay.
* Run template tests anyway.

Rollout
=======


cdmistman (approved):
🙌

PR #223, opened 2024-01-03
[python] unset PYTHONNOUSERSITE 

ryantm:
Why
===
* We set POETRY_USE_USER_SITE = 1 but some nixpkgs pkgs use a python
setup hook that sets PYTHONNOUSERSITE=1, which breaks poetry
installs. Reproduce breakage with:

```
nix-editor --add pkgs.magic-wormhole
poetry add pycparser
```

What changed
===
* In the python wrapper, unset PYTHONNOUSERSITE.

Test plan
===
```
nix-editor --add pkgs.magic-wormhole
poetry add pycparser
```
observe the poetry package installs fine

Rollout
=======


blast-hardcheese (approved):
good bandaid

PR #224, opened 2024-01-03
[ci] change dev disk path

ryantm:
Why
===
* The older generations were getting deleted, causing issues. Instead, lets upload them based on the rev, like we do for the production disks.

What changed
===
* renamed some variables from squashfs to dev to make it more clear what's going on
* update dev disk path to include the rev and be in a separate folder

Test plan
===
* run .scripts/build_disk_image.sh

Rollout
=======


masad-frost (approved):
:envelope:

PR #226, opened 2024-01-05
bug/Poetry env isolation

blast-hardcheese:
Why
===

Resolves [installing redberry breaks poetry](https://linear.app/replit/issue/DX-380/installing-redberry-breaks-poetry), as well as https://ask.replit.com/t/nix-error-when-installing-or-uninstalling-packages/94756

What changed
============

poetry wrapper removes `PYTHONUSERPATH` from `sys.path` before invoking `main()` to ensure that no user packages conflict with Poetry deps.

Test plan
=========

- [x] Already tested on devvm, verified bug, verified fix.
- [ ] Test in canary, install `redberry` (depends on `html5lib===1.0b8`, which causes the problem). Attempt to uninstall `redberry`, verify poetry does not crash.

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -547,6 +547,10 @@
     "commit": "2da491110c9866a9afb0bb9f668a6d35a8b98b43",
     "path": "/nix/store/lpn8jr69z6rbmhwmnf36y89lkbin3hrz-replit-module-python-3.10"
   },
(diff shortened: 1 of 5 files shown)


airportyh (approved):
Thanks!

PR #236, opened 2024-01-17
Supply nix cache secrets to builder

blast-hardcheese:
Why
===

So, this error:
image

seems to be due to `scripts/setup_nix_binary_cache.py`'s `create_netrc()` needing [these envvars](https://github.com/replit/nixmodules/blob/bb8c970212b268195022dc4b050bfc25d580f648/scripts/setup_nix_binary_cache.py#L49-L53)
image

which would be provided by

image

if this secret were available

image

What changed
============

Expose that secret to the build phase

Test plan
=========

Verify that warning disappears. Presumably the build cache being present will only make the build faster, and we can rely on nix to not ship something broken beyond that.

Rollout
=======

N/A

### DIFF
--- a/.semaphore/build-nixmodules.yml
+++ b/.semaphore/build-nixmodules.yml
@@ -40,6 +40,8 @@ blocks:
             - set -o pipefail
             - scripts/setup_nix_binary_cache.py
             - scripts/build_disk_image.sh 2>&1 | tee nixmodules_build.log
+      secrets:
+        - name: nix-build-cache-auth
       epilogue:
         always:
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
As I was requesting Ryan, GitHub's UI got excited and re-added devex, which then re-triggered the round-robbin, hence the PR party. Please pardon the noise.

ryantm (approved):

blast-hardcheese:
Looks like it shaved off ~15m from the build time 🎉

PR #237, opened 2024-01-18
Log netrc path and hash contents

blast-hardcheese:
Why
===

Despite #236, we're still getting the same error. Instead of guessing, let's see what the state of the world looks like.

What changed
============

Informational logging around the existence and path of what nix thinks the netrc file looks like. This should be safe to stay to help debug future config drift as well.

Test plan
=========

Tested `nix --extra-experimental-features nix-command show-config netrc-file` on a devbox

Rollout
=======

N/a

### DIFF
--- a/scripts/build_disk_image.sh
+++ b/scripts/build_disk_image.sh
@@ -10,6 +10,13 @@ set -eo pipefail
 # To run, make sure you are logged in via:
 # gcloud auth login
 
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
Not necessary, `sem debug job ...` allowed me to get access to the environment and confirm the issue.

PR #238, opened 2024-01-18
Supplying poetry config.toml with empty contents to suppress poetry error

blast-hardcheese:
Why
===

Resolves https://linear.app/replit/issue/DX-367/poetry-config-errors .

`poetry config` attempts to create the config file at `$POETRY_CONFIG_DIR/config.toml` if it doesn't exist. An empty file is fine.

The previous `conf.toml` does not appear to be used by anything.

What changed
============

Removed `conf.toml`, created an empty `config.toml`

Test plan
=========

Tested `poetry conf` locally with an empty `config.toml`

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -28,13 +28,8 @@ let
   poetry-config = pkgs.writeTextFile {
     name = "poetry-config";
     text = ''
-      [[tool.poetry.source]]
-      name = "replit"
-      url = "https://package-proxy.replit.com/pypi/simple/"
-      default = true
-
     '';
-    destination = "/conf.toml";
+    destination = "/config.toml";
   };
 
   debugpy = pypkgs.debugpy.overridePythonAttrs
--- a/modules.json
+++ b/modules.json
@@ -595,6 +595,10 @@
     "commit": "0bd73cd137ee828d927e95d97e8199e7b66b6c29",
     "path": "/nix/store/ii8s3313impgc8v74l9v9nvr85lvmzng-replit-module-python-3.10"
   },
+  "python-3.10:v41-20240117-c740790": {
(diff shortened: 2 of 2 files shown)


pkgs/modules/python/default.nix
```diff
@@ -25,16 +25,23 @@ let
...
+  #  poetry-config = pkgs.writeTextFile {
+  #    name = "poetry-config";
+  #    text = ''
+  #      [[tool.poetry.source]]
+  #      name = "replit"
+  #      url = "https://package-proxy.replit.com/pypi/simple/"
+  #      default = true
+  #
+  #    '';
+  #    destination = "/conf.toml";
```
ryantm, line 37: I don't think we should leave this here commented out, because it might trick people into thinking conf.toml does something.

ryantm (changes requested):
Do we even need the POETRY_CONFIG_DIR environment variable if we aren't doing anything with the file?

blast-hardcheese:
> Do we even need the POETRY_CONFIG_DIR environment variable if we aren't doing anything with the file?

I'm inclined to set it, to reserve its existence for when we need to influence poetry's behavior.

It defaults to `XDG_CONFIG_DIR/pypoetry`, which is `.config` inside the repl working directory, so it wouldn't even be ephemeral. I'm not convinced most users will need to do anything with that, looking at [available settings](https://python-poetry.org/docs/configuration/#available-settings), though it has quite a few settings that we will likely need in the not-too-distant future.

pkgs/modules/python/default.nix
blast-hardcheese: 👍

The contents are effectively the same as what's in our pip wrapper, so if needed they can be pulled out of git history or that file.

ryantm (approved):
Okay sounds good.

airportyh (approved):

PR #241, opened 2024-01-23
:hocho: remove non-rg CI/CD

ryanwe:
Why
===

We no longer run repls in marine-cycle.

This is the first step. After we're no longer deploying code, we can begin to 🪓 the physical infra.

See https://github.com/replit/goval/pull/12109 for other services

What changed
============

remove non-rg deployments.

Test plan
=========

Rollout
=======


ryantm (approved):

manifests/application.pvn.yaml
```diff
@@ -4,17 +4,24 @@ application:
...
+    - name: staging
       constants:
         - name: name
           string:
-            value: canary
+            value: staging
+      runtimes:
+        - name: config-raw
+          runtime: govalctl-config-raw
+          type: EXTENSION
```
cdmistman, line 15: what's the status of staging? i thought we're only doing prerelease for now?

ryanwe: Staging still exists but is mostly just for staging infra changes, we don't run repls there. Eventually, we want an end-to-end staging environment that is separate from prod. But web isn't able to support multiple environments yet.

ryanwe:
Closing since these files were already removed.

PR #242, opened 2024-01-23
use dotreplit schema from repl environment

cdmistman:
Why
===

now that we're auto-generating the JSON schema in goval, time to use that auto-generated schema :)

What changed
============

- removed `dotreplit-schema.json`
- changed `replit` nixmodule to use `/etc/replit/dotreplit.schema.json`

Test plan
=========

.replit lsp works

Rollout
=======


blast-hardcheese (approved):
👀

PR #246, opened 2024-01-26
DX-415: Temporarily work around upm lock slowdown

blast-hardcheese:
Why
===

upm calls out to `poetry env list` during guess, add, lock, and remove. Inside a repl, poetry isn't (typically?) configured in such a way for this to return any entries, so just suppress that subcommand until `.pythonlibs` gets promoted to a virtualenv.

The only caveat here is if users have explicitly altered their envvars such that poetry is installing to a venv somewhere inside the repl directory, this would disable one of the conditionals in upm that would trigger a call to `Lock`.

What changed
============

Alter the `poetry-in-venv` wrapper to make any arglist starting with `env list` just `exit 0`, bypassing poetry.

Test plan
=========

I copied the poetry wrapper inside a repl, added these, and modified `poetry.lock` to trigger upm's desire to recheck the lock, verifying the early termination and subsequent speedup.

Rollout
=======

I don't anticipate this workaround being needed for more than a month, but should improve UX for users during that time.

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -599,6 +599,10 @@
     "commit": "c740790111c4166c1764ba7073662129eef6b9d6",
     "path": "/nix/store/yl6qynci35jjjw50h7ig1sq8c78a8qay-replit-module-python-3.10"
   },
+  "python-3.10:v42-20240125-d91747b": {
+    "commit": "d91747b1a9655b307a69613063d0599aef8b8f3e",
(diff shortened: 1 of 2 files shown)


blast-hardcheese:
I looked into why this was slow in the first place as well. There's a lot that needs to be initialized to get the answer, even `poetry env info` takes twice as long as `poetry env list`. There's no "just get me the path to the current venv".

ryantm (approved):

PR #252, opened 2024-02-02
DX-427: Work around for /repl/stats not existing in Deployments

blast-hardcheese:
Why
===

Using `poetry` inside Deployments' `run` command tries (and fails) to access `/repl/...`, which has been asserted to never exist.

What changed
============

Replicated the logic that emits that field in `resources.json`, delegating to `jq` to get floating-point division.

Test plan
=========

Copied this modified script into a repl and used it in the following command: `./poetry.sh run python -m http.server`
With plain `poetry` I can reproduce the error, with this change the error is resolved.

Rollout
=======

### DIFF
--- a/pkgs/poetry/poetry-in-venv.nix
+++ b/pkgs/poetry/poetry-in-venv.nix
@@ -24,10 +24,24 @@ in
 pkgs.writeShellApplication {
   name = "poetry";
   text = ''
+    # Determine how many vcpus we have, first in the standard location, with a fallback
+    # to the cgroup info, since resources.json does not exist during Deployments.
+    if [ -e /repl/stats/resources.json ]; then  # Common resources location
+      numVCpu="$(${pkgs.jq}/bin/jq .numVCpu </repl/stats/resources.json)"
+    else
+      cgroup_path="$(cut -f 3 -d : < /proc/1/cgroup)"
+      if [ -f "$cgroup_path" ]; then
+        cpu_path="/sys/fs/cgroup/$cgroup_path/cpu.max"
+        if [ -f "$cpu_path" ]; then
+          numVCpu="$(${pkgs.jq}/bin/jq -R 'split(" ") | map(tonumber) | .[0] / .[1]' < "$cpu_path")"
+        fi
+      fi
+    fi
+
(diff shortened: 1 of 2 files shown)


pkgs/poetry/poetry-in-venv.nix
```diff
@@ -24,12 +24,29 @@ in
...
-    # Don't run multiple install workers in parallel if we have only 0.5 CPUs:
-    # helps with speed in free accounts
-    numVCpu=$(${pkgs.jq}/bin/jq .numVCpu </repl/stats/resources.json)
-    if [ "''${numVCpu}" = "0.5" ]; then
+    # Determine how many vcpus we have, first in the standard location, with a fallback
+    # to the cgroup info, since resources.json does not exist during Deployments.
+    if [ -e /repl/stats/resources.json ]; then  # Common resources location
+      numVCpu="$(${pkgs.jq}/bin/jq .numVCpu </repl/stats/resources.json)"
+    else
+      cgroup_path="$(cat /proc/1/cgroup | cut -f 3 -d :)"
```
ryantm, line 32: Did you try `cat /proc/1/cgroup` in both Autoscale and Reserved Deployments?

blast-hardcheese: Only in Autoscale, I'll give `Reserved` a try, but I was operating under the assumption that we'd always have cgroup limitations, subject to approval by @lincoln-replit

airportyh (approved):

pkgs/poetry/poetry-in-venv.nix
lincoln-replit: I can't think of any reason why it shouldn't work but it would be good to verify that this command works in reserved VMs as well!

blast-hardcheese: Pardon the late follow-up. Turns out I am, in fact, falling back to defaults (`0.5` vcpus) due to pid1's cgroup file not existing. I may be misremembering, but I think I heard someone say we just flat out do not run pid1 inside deployment contexts. Is that true? I will keep digging when I come back to this

ryantm (approved):

PR #253, opened 2024-02-02
Add community-version field

airportyh:
Why
===

Add community version (ex. python full version) to the JSON for display purpose.


pkgs/moduleit/module-definition.nix
```diff
@@ -570,6 +570,11 @@ in
       description = "Name of the module";
     };
 
+    community-version = mkOption {
```
ryantm, line 573: How about communityVersion? I think that matches the other options.

PR #254, opened 2024-02-02
chore/atty dependabot warning

blast-hardcheese:
Why
===

Dependabot suggests that `atty` has some unlikely vulnerability running in an environment we don't support, but there's a newer version of `clap` that moved away from using `atty` (which is otherwise abandoned) anyway, so just use it.

What changed
============

Bump from clap 3.1 to 4.0, `cargo update`

Test plan
=========

I ran the binary locally and it got past CLI arg parsing, which I presume is sufficient.

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -1135,6 +1135,10 @@
     "commit": "d91747b1a9655b307a69613063d0599aef8b8f3e",
     "path": "/nix/store/58p275vnh2ajp0bi3w43pbj0150w39jk-replit-module-python-with-prybar-3.10"
   },
+  "python-with-prybar-3.10:v22-20240203-ce85ddb": {
+    "commit": "ce85ddbaacf5bb78c6fec1174ad6899abe102850",
+    "path": "/nix/store/brnql4lydsahvv3xhw445nn544cq80fs-replit-module-python-with-prybar-3.10"
+  },
   "nodejs-with-prybar-18:v1-20230928-1ea5ac0": {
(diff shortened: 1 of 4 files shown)


ryantm (approved):
Thanks for staying on top of these.

blast-hardcheese:
I had to constrain the `clap` version to something compatible with rustc 1.69.0 which is what I presume is available in `pkgs-23_05`, at least until I can fix Python to build with unstable like everything else.

modules.json
```diff
@@ -1135,6 +1135,14 @@
...
     "path": "/nix/store/58p275vnh2ajp0bi3w43pbj0150w39jk-replit-module-python-with-prybar-3.10"
   },
+  "python-with-prybar-3.10:v22-20240203-c155d94": {
+    "commit": "c155d94129642c7241ade447ce8a039e79114100",
+    "path": "/nix/store/9smzvywljsywi1xpxx9zgdj4ywbsywx0-replit-module-python-with-prybar-3.10"
+  },
+  "python-with-prybar-3.10:v23-20240203-f1b2c8e": {
+    "commit": "f1b2c8e5055494db72f4b471d59275487f0e17a7",
+    "path": "/nix/store/0sf38jckc3valaxhfw666l94yhm3pfwd-replit-module-python-with-prybar-3.10"
+  },
```
cdmistman, line 1145: please only 1 release!

PR #256, opened 2024-02-06
chore/liberate python

blast-hardcheese:
Why
===

Switch python back to `nixpkgs-unstable`, `bootstrapped-pip` has been a noop for a while, from what I can tell (see https://github.com/NixOS/nixpkgs/pull/251824 for a little bit of context, though the implications are still somewhat over my head.)

Presumably if the disk builds and template tests pass, we're good to go.

What changed
============

Deleted the reference to bootstrapped-pip, switched python over to `nixpkgs-unstable`

Test plan
=========

- Built the disk locally with all suppported versions of python
- Ran some tests against local conman

Rollout
=======

### DIFF
--- a/pkgs/pip/default.nix
+++ b/pkgs/pip/default.nix
@@ -26,7 +26,7 @@ pypkgs.buildPythonPackage rec {
     name = "${pname}-${version}-source";
   };
 
-  nativeBuildInputs = [ pypkgs.bootstrapped-pip ];
+  nativeBuildInputs = [ ];
 
   # pip detects that we already have bootstrapped_pip "installed", so we need
   # to force it a little.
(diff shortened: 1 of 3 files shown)


pkgs/pip/default.nix
```diff
@@ -26,7 +26,7 @@ pypkgs.buildPythonPackage rec {
     name = "${pname}-${version}-source";
   };
 
-  nativeBuildInputs = [ pypkgs.bootstrapped-pip ];
+  nativeBuildInputs = [ ];
 
   # pip detects that we already have bootstrapped_pip "installed", so we need
```
ryantm, line 31: There's still a mention of bootstrapped_pip here.

ryantm (approved):
Looks reasonable. What prompted the change? Just looking into what bootstrapped-pip was?

blast-hardcheese:
@ryantm Colton couldn't upgrade to unstable nixpkgs due to the unavailability of bootstrapped-pip, I figured it was better to solve this sooner rather than have the Python modules succumb to bitrot

blast-hardcheese:
Closing due to #257 including this changeset

PR #257, opened 2024-02-06
Update python to unstable channel

airportyh:
Why
===

We have a lot of instance of people updating their nix channel in .replit to stable-23_11 and then sys deps causing this to break. If python's nix channel is at or ahead of stable-23_11, which breakage will be far less common. This is a short-term fix before we have a good rtld-loader-based solution.

What changed
============

1. Updated python-3.10 and 3.11's nix channel to unstable (which is ahead of 23_11)
2. Revert pip to pypkgs.pip
3. added pkgs-unstable argument for convenience when you need to pick between the 2 channels

Test plan
=========

1. template tests and sanity checks

### DIFF
--- a/pkgs/moduleit/entrypoint.nix
+++ b/pkgs/moduleit/entrypoint.nix
@@ -11,6 +11,7 @@ let
     ];
     specialArgs = {
       inherit pkgs pkgs-23_05;
+      pkgs-unstable = pkgs;
       modulesPath = builtins.toString ./.;
     };
   });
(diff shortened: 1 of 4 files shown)


pkgs/moduleit/entrypoint.nix
```diff
@@ -11,6 +11,7 @@ let
     ];
     specialArgs = {
       inherit pkgs pkgs-23_05;
+      pkgs-unstable = pkgs;
```
ryantm, line 14: Why do we need to send pkgs in twice (once as pkgs and once as pkgs-unstable)?

airportyh: That's just so we can have in under a different name than pkgs, because I want to use pkgs internally as a local var by picking one from pkgs-unstable vs pkgs-23_05.

ryantm: Okay, yeah it looks like there is no way around this aside from renaming the pkgs inside the python module.

ryantm (approved):

PR #258, opened 2024-02-06
[chore] Also upgrade python 3.8

blast-hardcheese:
Why
===

#256 had 3.8, but #257 skipped it. Tested locally, it seems to work, so upgrading that as well.

What changed
============

python 3.8 is no longer special-cased to 23_05.

Test plan
=========

Install python 3.8 module, try to import libraries that use native `.so` files

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -3,7 +3,7 @@
 let
   pythonVersion = lib.versions.majorMinor python.version;
 
-  pkgs = if pythonVersion == "3.8" then pkgs-23_05 else pkgs-unstable;
+  pkgs = pkgs-unstable;
 
   pylibs-dir = ".pythonlibs";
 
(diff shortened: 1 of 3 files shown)


pkgs/modules/python/default.nix
```diff
@@ -3,7 +3,7 @@
 let
   pythonVersion = lib.versions.majorMinor python.version;
 
-  pkgs = if pythonVersion == "3.8" then pkgs-23_05 else pkgs-unstable;
+  pkgs = pkgs-unstable;
```
ryantm, line 6: If we're switching to pkgs-unstable, we can undo this change I was slightly complaining about here too and just reference pkgs directly: https://github.com/replit/nixmodules/pull/257/files#r1480135676

ryantm (changes requested):

airportyh:
Hold this...gonna need to [revert the python upgrade until rtld loader lands](https://replit.slack.com/archives/C03KS2B221W/p1707315334964009) :(

blast-hardcheese:
👍 Going to punt on this change for now due to observations. In Slack, it's determined that we'll likely need to maintain 3.8 on 23_05 until we delete it.

PR #264, opened 2024-02-13
bun: support replit ld library path

cdmistman:
Why
===

user asked for it

What changed
============

1. made an abstraction over the node wrapper generator for `REPLIT_LD_LIBRARY_PATH`
2. changed wrapped node to use abstraction
3. used abstraction for bun

Test plan
=========

with the bun module installed:
- in `.replit` set `REPLIT_LD_LIBRARY_PATH = "foo"`
- change `index.ts` to be `console.log(process.env.LD_LIBRARY_PATH)`
- running the program via run button results in `foo` being present
- running the program via `bun index.ts` has the same result

Rollout
=======

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -37,6 +37,35 @@
     {
       overlays.default = final: prev: {
         moduleit = self.packages.${prev.system}.moduleit;
+
+        lib = prev.lib // {
+          mkWrapper-replit_ld_library_path = package: final.stdenvNoCC.mkDerivation {
+            name = "${package.name}-wrapped";
+            inherit (package) meta version;
+
+            buildInputs = [ pkgs.makeWrapper ];
+            buildCommand = ''
+              if [ ! -d ${package}/bin ]; then
+                echo "No bin directory found in ${package}"
+                exit 1
+              fi
+          
+              mkdir -p $out/bin
+          
+              for bin in ${package}/bin/*; do
+                local binName=$(basename $bin)
+                cat >$out/bin/$binName <<-EOF
+              #!${final.bash}/bin/bash
+              if [ -n "\''${REPLIT_LD_LIBRARY_PATH-}" ]; then
+                export LD_LIBRARY_PATH="\$REPLIT_LD_LIBRARY_PATH:\$LD_LIBRARY_PATH"
+              fi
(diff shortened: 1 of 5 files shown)


pkgs/modules/nodejs/default.nix
```diff
@@ -2,26 +2,27 @@
...
+  #   buildCommand = ''
+  #     mkdir -p $out/bin
+  #     for bin in ${nodejs}/bin/*; do
+  #       local binName=$(basename $bin)
+  #       cat >$out/bin/$binName <<-EOF
+  #     #!${pkgs.bash}/bin/bash
+  #     if [ -n "\''${REPLIT_LD_LIBRARY_PATH-}" ]; then
+  #       export LD_LIBRARY_PATH="\$REPLIT_LD_LIBRARY_PATH:\$LD_LIBRARY_PATH"
+  #     fi
+  #     exec "$bin" "\$@"
```
blast-hardcheese, line 17: just delete the comment?

flake.nix
```diff
@@ -37,6 +37,30 @@
...
+        lib = prev.lib // {
+          mkWrapper-replit_ld_library_path = package: final.stdenvNoCC.mkDerivation {
+            name = "${package.name}-wrapped";
+            inherit (package) meta version;
+
+            buildInputs = [ pkgs.makeWrapper ];
+            buildCommand = ''
+                            mkdir -p $out/bin
+
+                            for bin in ${package}/bin/*; do
```
blast-hardcheese, line 50: so, one annoyance is that if there _is_ no `bin` in a package, this'll just be the literal string specified here:
image

probably not a huge deal, but considering you're learning shell, it seems prudent to mention.

Maybe throw in a `[ ! -d ${package}/bin ] || exit 0`, or a `bins=( ${package}/bin/* )`/`[ -f ${bins[0]} ] || exit 0`/`for bin in "${bins[@]}"; do`

blast-hardcheese (approved):
looks sensible

flake.nix
blast-hardcheese: 🙏

blast-hardcheese (approved):

PR #265, opened 2024-02-14
Python to use rtld loader

airportyh:
Why
===

Use RTLD loader instead of LD_LIBRARY_PATH to avoid glibc and other shared library version mismatch issues.

Depends on https://github.com/replit/goval/pull/12350

What changed
============

1. if the pid1-replit-rtld-loader flag is on, and the Nix channel in .replit is not the legacy channel (we know the glibc in that version crashes with LD_AUDIT), then enable RTLD loader in lieu of setting LD_LIBRARY_PATH

Test plan
=========

Revisit scenarios 1, 2, 4 in [this doc](https://docs.google.com/document/d/1llRzZdBZIKDFk5n5NQromYMCeDaYUCB9pVLy2vahTH4) and see that they are fixed. When https://github.com/replit/goval/pull/12312 is in and Python in nixmodules is upgraded to latest Nix channel, all 5 should be fixed.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -252,6 +252,26 @@
         "type": "github"
       }
     },
+    "replit-rtld-loader": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1707923725,
+        "narHash": "sha256-FkahFyXa4OAE18HZz47da9IVCbV3t8EyClaVTvCWEkk=",
+        "owner": "replit",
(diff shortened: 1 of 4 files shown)


blast-hardcheese (approved):
What sets `REPLIT_RTLD_LOADER`? Currently just `replit.nix`?

airportyh:
> What sets `REPLIT_RTLD_LOADER`? Currently just `replit.nix`?

That's in a [pid1 PR](https://github.com/replit/goval/pull/12350).

PR #266, opened 2024-02-15
[python] Delete temporary poetry workaround

blast-hardcheese:
Why
===

This performance workaround is no longer needed

What changed
============

- Workaround was codified in https://github.com/replit/upm/pull/230/
- Recent work to make `VIRTUAL_ENV` set for Python modules will add an additional buffer before running this command

Test plan
=========

Run `upm show-package-dir`, see whether it runs `poetry env list --full-path`

Rollout
=======


ryantm (approved):
Thanks for cleaning this up.

PR #270, opened 2024-02-26
Update python Nix channel (attempt 2)

airportyh:
Why
===

Second attempt to update Nix channel to unstable for Python modules ([first attempt](https://github.com/replit/nixmodules/pull/259)). Will do it with a more minimal change this time: with a small tweak to pip (removing the buildInput, which was tested to work in dev). Also, it shouldn't cause breakage of [this kind](https://replit.slack.com/archives/C03KS2B221W/p1707315334964009) now that rtld loader has been deployed.

What changed
============

1. updated pkgs from 23_05 to unstable for all Python modules
2. removed `pypkgs.bootstrapped-pip` from pip as nativeBuildInput

Test plan
=========

Run the template tests in prerelease and see that python template tests still pass.

### DIFF
--- a/pkgs/modules/default.nix
+++ b/pkgs/modules/default.nix
@@ -12,16 +12,16 @@ let
 
   modulesList = [
     (import ./python {
-      python = pkgs-23_05.python38Full;
-      pypkgs = pkgs-23_05.python38Packages;
+      python = pkgs.python38Full;
+      pypkgs = pkgs.python38Packages;
     })
     (import ./python {
-      python = pkgs-23_05.python310Full;
-      pypkgs = pkgs-23_05.python310Packages;
+      python = pkgs.python310Full;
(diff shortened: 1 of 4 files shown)


pkgs/modules/default.nix
```diff
@@ -12,16 +12,16 @@ let
 
   modulesList = [
     (import ./python {
-      python = pkgs-23_05.python38Full;
```
blast-hardcheese, line 15: won't 3.8 still fail to build `python38Packages` because sphinx is unsupported now? I think we're stuck unless we try to patch

airportyh: It worked for me. CI will test the build too.

blast-hardcheese: Huh. OK!

blast-hardcheese (approved):

PR #272, opened 2024-02-29
Update nix channel for python; use stock pip; monkeypatch pip to chan…

airportyh:
…ge shebang line

Why
===

1. We want to upgrade [Python to nixpkgs-unstable ](https://linear.app/replit/issue/DX-466/upgrade-python-in-nix-modules-to-nixpkgs-unstable)
2. due to an [update in nixpkgs,](https://github.com/NixOS/nixpkgs/pull/245509) we had to change how we package pip, we chose to:
    a. use the stock pip instead of our custom one
    b. add back a customization for the shebang line via monkeypatching, via [python site customization](https://docs.python.org/3/library/site.html), which has the advantage of working for a pip that's been installed in .pythonlibs. Thanks @blast-hardcheese for this!

What changed
============

1. Use unstable channel for Python 3.10 and 3.11, but not 3.8.
2. Updated pip to use stock pip, plus site customization
3. set `break-system-packages = yes` to silence [this error](https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3)

Test plan
=========

1. create a Python Repl
2. `pip install gunicorn`
4. `cat .pythonlibs/bin/gunicorn` see that the shebang line starts with `#!/usr/bin/env python3`
5. make sure template tests (in particular pip and poetry tests) pass

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -711,6 +711,10 @@
     "commit": "aba8eb66f6ff96c4f341bb18d71c715a2501ec83",
     "path": "/nix/store/55d7s1ydlwfbv03sfyd38vvszp16ddwk-replit-module-python-3.10"
   },
+  "python-3.10:v50-20240229-2250a9a": {
+    "commit": "2250a9afde519540b54c69522bc27f732654f294",
+    "path": "/nix/store/i0q9y4kjp4lz216n6qfkl9vs96arxjq2-replit-module-python-3.10"
+  },
   "qbasic:v1-20230525-c48c43c": {
     "commit": "c48c43c6c698223ed3ce2abc5a2d708735a77d5b",
     "path": "/nix/store/4n4raazspqy2zgawkkyaqc9xapl1f5sz-replit-module-qbasic"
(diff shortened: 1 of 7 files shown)


airportyh:
@blast-hardcheese I think we should expose `PIP_CONFIG_FILE` var to the repl so that an upgraded pip will have `break-system-packages = yes` set as well.

airportyh:
@blast-hardcheese I think we want to make the script not break if for some reason pip was not available on PYTHONPATH.

airportyh:
I can't find a way to break it by setting the PYTHONPATH though, so it ok.

ryantm:
I'm concerned about the test plan here, because it already does what the test plan says without this change.

blast-hardcheese:
@ryantm That won't be true in the upgraded nixmodules -- that is the case today, and our goal is to effectively have a noop through this upgrade

blast-hardcheese (approved):

ryantm (approved):

PR #275, opened 2024-03-01
Fix site customize 2

airportyh:
Why
===

Oops, I merged the https://github.com/replit/nixmodules/pull/272 before testing. sitecustomize.py broke: it contained the path to the file rather than the contents. This fixes it (version 2).

## Test Plan
1. create a Python Repl
2. pip install gunicorn
3. cat .pythonlibs/bin/gunicorn see that the shebang line starts with #!/usr/bin/env python3
4. make sure template tests (in particular pip and poetry tests) pass
5. `pip install --upgrade pip`
6. `which pip` to see we are using the pip inside .pythonlibs
7. try using that pip to install a thing: `pip install flask`

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -715,6 +715,10 @@
     "commit": "2250a9afde519540b54c69522bc27f732654f294",
     "path": "/nix/store/i0q9y4kjp4lz216n6qfkl9vs96arxjq2-replit-module-python-3.10"
   },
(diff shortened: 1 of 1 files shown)


airportyh:
Verified the shebang line works even with the upgraded version of pip.

ryantm (approved):

PR #277, opened 2024-03-04
Th sitecustomize subdir

airportyh:
Why
===

We found that the sitecustomize.py setup we added makes pyright hang when starting up, and we had failed lsp tests. 
https://replit.slack.com/archives/C06KAKG35QS/p1709318856982249?thread_ts=1709316270.385049&cid=C06KAKG35QS

What changed
============

Moved sitecustomize.py to a subdirectory `lib/python/site-packages` to prevent pyright from trying to walk up and list the files in `/nix/store`

Test plan
=========

Same as in https://github.com/replit/nixmodules/pull/275

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -719,6 +719,10 @@
     "commit": "fc137ca591a3c0e819099783a09b6ba567c1f246",
     "path": "/nix/store/z7v403js03f8xhd3vpg41iwkzfgb5ahc-replit-module-python-3.10"
   },
+  "python-3.10:v52-20240304-83361d1": {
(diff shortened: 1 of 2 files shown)


ryantm (approved):
The single file vs folder nix output distinction strikes again. Nice job debugging!

airportyh:
> The single file vs folder nix output distinction strikes again. Nice job debugging!

It was already a file in a folder, but this makes it go in more nested folders. I believe it would be a single file wo a destination specified.

PR #279, opened 2024-03-06
Th more display versions

airportyh:
Why
===

We want to have display versions for all language servers, formatters, and packagers.

https://linear.app/replit/issue/DX-504/add-display-version-to-all-formatters-languageservers-and-packagers

What changed
============

Added reasonable display versions to those.

Test plan
=========

1. `nix develop`
2. `nix build .#registry`
3. `python scripts/show_reg.py`

### DIFF
--- a/pkgs/modules/go/default.nix
+++ b/pkgs/modules/go/default.nix
@@ -26,6 +26,7 @@ in
   replit.dev.formatters.go-fmt = {
     name = "go fmt";
     language = "go";
+    displayVersion = "Go ${go.version}";
 
     start = "${go}/bin/go fmt";
     stdin = false;
@@ -34,6 +35,8 @@ in
   replit.dev.languageServers.gopls = {
     name = "gopls";
     language = "go";
+
+    displayVersion = gopls.version;
 
     start = "${gopls}/bin/gopls";
   };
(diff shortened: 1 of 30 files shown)


pkgs/modules/go/default.nix
```diff
@@ -26,6 +26,7 @@ in
   replit.dev.formatters.go-fmt = {
     name = "go fmt";
     language = "go";
+    displayVersion = "Go ${go.version}";
```
bradymadden97, line 29: Why is this prefixed with "Go"?

pkgs/modules/ruby/default.nix
```diff
@@ -47,12 +47,15 @@ in
...
+
     start = "${rubyPackages.solargraph}/bin/solargraph stdio";
   };
 
-  replit.packagers.ruby = {
-    name = "Ruby";
+  replit.packagers.gem = {
+    name = "Gem";
     language = "ruby";
+    displayVersion = "Ruby ${ruby-version}";
```
bradymadden97, line 58: Same w/ this

bradymadden97 (approved):
Seems like some are prefixed w/ names and some not, curious why. But accepting to unblock

pkgs/modules/go/default.nix
airportyh: It's a "display version" so I used some leeway and put in what I wanted a user to see. But I think we are abandoning this approach in favor of going straight for Nixmodules V2.

PR #281, opened 2024-03-08
Introducing REPLIT_PYTHONPATH; avoid putting python's system libs in PYTHONPATH

airportyh:
Why
===

This fixes the python error:
```
    assert _sre.MAGIC == MAGIC, "SRE module mismatch"
           ^^^^^^^^^^^^^^^^^^^
AssertionError: SRE module mismatch
```
which occurs often when a user is using a Python Repl, and then adds a Nix dependency in replit.nix that is either a Python app or is another version of Python itself.
Exhibits:
* https://replit.slack.com/archives/C03KS2B221W/p1709838919631539
* https://replit.slack.com/archives/C03KS2B221W/p1709739288354329
* https://replit.slack.com/archives/C03KS2B221W/p1708711425824829

## Explanation

1. The standard regular expression library in Python `re` uses an internal library called `_sre` which is built-in to the Python interpreter. Both `re` and `_sre` contain in them a magic number, which `re` verifies are the same before any operations. If they are different, it fails loudly.
2. in [the python module we add the standard library path](https://github.com/replit/nixmodules/blob/main/pkgs/modules/python/default.nix#L157)  to `PYTHONPATH` ahead of `.pythonlibs/...` to prevent a pypi package of the same name from taking precedence over a standard lib, known example is [uuid](https://pypi.org/project/uuid/)
3. when you combine 1 and 2 and then try to run a different version of python from the one coming from the module, we get this error because `re` was found on `PYTHONPATH` via the standard lib path of "our" python, but the running python is different and therefore its `_sre` module has a different magic number from the loaded `re` library.

This problem is similar to the LD_LIBRARY_PATH problem: the precedence is wrong. Python already has its own standard libs in `sys.path`, we only added it to prevent `.pythonlibs` from having higher precedence to it. What we want is to add `.pythonlibs` to `PYTHONPATH` but in a way that has lower precedence than what Python already has setup by default.

What changed
============

1. Added support for `REPLIT_PYTHONPATH` in sitecustomize.py
2. Use PYTHONPATH only to deliver sitecustomize.py; use `REPLIT_PYTHONPATH` for everything else
3. Stop putting the python standard lib in the path, Python already has it by default, now only put:
   * `.pythonlibs/...`
   * pip's lib path so users can still do `python -m pip install ...`
   * setuptools lib path so certain libraries depending on it as a post install step can work (I am still not sure this is the right thing to do)
5. Move the PYTHONPATH usage in python-utils out to top level but convert to `REPLIT_PYTHONPATH`

Test plan
=========

1. Create a python repl
2. Add pkgs.magic-wormhole
3. run `wormhole` and it should work
4. Try the same with pkgs.gcloud
6. Adde pkgs.python312 (from unstable)
7. run python3.12 and it should work

Regression test with template tester.

### DIFF
--- a/pkgs/modules/python/sitecustomize.py
+++ b/pkgs/modules/python/sitecustomize.py
@@ -25,3 +25,9 @@ exe = sys.argv[0]
 if exe.endswith('/pip') or exe.endswith('/pip3') or exe.endswith('/.pip-wrapped'):
     from pip._vendor.distlib.scripts import ScriptMaker
     ScriptMaker._build_shebang = pip_build_shebang
+
+import os
+pythonpath = os.getenv('REPLIT_PYTHONPATH')
+if pythonpath:
+  paths = pythonpath.split(':')
+  sys.path.extend(paths)
--- a/modules.json
+++ b/modules.json
@@ -727,6 +727,10 @@
     "commit": "46fe79374aa5040f53b356290b471547ac2dc45c",
     "path": "/nix/store/njkmnnqvaqq4cvg01s4zzl4lza7420qd-replit-module-python-3.10"
   },
+  "python-3.10:v54-20240312-53a08a2": {
+    "commit": "53a08a23d7154937437e54e52158f5f948887e63",
+    "path": "/nix/store/1ivb7i0aqwj0yqzlfmwl1g3dm5w836wa-replit-module-python-3.10"
+  },
   "qbasic:v1-20230525-c48c43c": {
     "commit": "c48c43c6c698223ed3ce2abc5a2d708735a77d5b",
     "path": "/nix/store/4n4raazspqy2zgawkkyaqc9xapl1f5sz-replit-module-qbasic"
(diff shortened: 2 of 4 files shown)


pkgs/modules/python/sitecustomize.py
```diff
@@ -25,3 +25,9 @@ def pip_build_shebang(self, executable, post_interp):
 if exe.endswith('/pip') or exe.endswith('/pip3') or exe.endswith('/.pip-wrapped'):
     from pip._vendor.distlib.scripts import ScriptMaker
     ScriptMaker._build_shebang = pip_build_shebang
+
+import os
+pythonpath = os.getenv('REPLIT_PYTHONPATH')
+if pythonpath:
+  paths = pythonpath.split(':')
+  sys.path.extend(paths)
```
blast-hardcheese, line 33: This would prevent users from ever overriding anything that's in the standard path, we should instead
```suggestion
  paths.extend(sys.path)
  sys.path = paths
```

blast-hardcheese: Not correct!

blast-hardcheese (approved):

airportyh:
Tested in local. Works intended. But, it doesn't mesh as well as python/magic-wormhole from nix as I hoped because those Nix packages export `PYTHONPATH` entries which prepend ahead of our sitecustomization :) But that's something to solve for another day.

PR #285, opened 2024-03-15
Revert "Revert "Introducing REPLIT_PYTHONPATH; avoid putting python's system libs in PYTHONPATH""

airportyh:
Reverts replit/nixmodules#283. Re-lands https://github.com/replit/nixmodules/pull/281 introducing REPLIT_PYTHONPATH. It broke pip/poetry because pip was moved out of PYTHONPATH causing sitecustomize not to be able to find it. This moves pip back, but not the others.

## Test

Same procedure as in https://github.com/replit/nixmodules/pull/281


blast-hardcheese (approved):
Nice!

PR #287, opened 2024-03-18
V2 setup

airportyh:
Why
===

Start conversion to V2 style modules. See [more here](https://docs.google.com/document/d/1Khv0ZNCPaWWma6Oh3Q7P5KMI4JnNPSpejcNVoaEDLYc).

What changed
============

Some preliminary setup for making the changes.

Test plan
=========

`nix eval .#allModulesOptions --json` should give an empty object.

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -148,7 +148,7 @@ rec {
       specialArgs = {
         inherit pkgs pkgs-23_05;
         pkgs-unstable = pkgs;
-        modulesPath = builtins.toString ./.;
+        modulesPath = ./modules;
       };
     });
 
@@ -164,7 +164,7 @@ rec {
       specialArgs = {
         inherit pkgs pkgs-23_05;
         pkgs-unstable = pkgs;
-        modulesPath = builtins.toString ./.;
+        modulesPath = ./modules;
       };
     });
     lib = pkgs.lib;
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -1,4 +1,4 @@
-{ pkgs, self }:
+{ pkgs, pkgs-23_05, self }:
 
 with pkgs.lib;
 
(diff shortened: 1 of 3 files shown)


pkgs/default.nix
```diff
@@ -116,4 +116,63 @@ rec {
...
+    # (import ./modules/formatters/prettier)
+    # (import ./modules/languageServers/typescript-language-server)
+    # (import ./modules/packagers/nodejs-packager)
+    # (import ./modules/debuggers/js-debug)
+    # (import ./modules/bundles/bun)
+    # (import ./modules/interpreters/bun)
+    # (import ./modules/packagers/bun)
+    # (import ./modules/bundles/web)
+    # (import ./modules/languageServers/css-language-server)
+    # (import ./modules/languageServers/html-language-server)
```
cdmistman, line 140:
```suggestion
```

the dir structure of this repo has really been bothering me, let's fix that with this new era 😅

pkgs/default.nix
```diff
@@ -116,4 +116,63 @@ rec {
...
+
+  myEvalModule = path:
+    (pkgs.lib.evalModules {
+      modules = [
+        (import path)
+      ] ++ allModules;
+      specialArgs = {
+        inherit pkgs pkgs-23_05;
+        pkgs-unstable = pkgs;
+        modulesPath = builtins.toString ./.;
```
cdmistman, line 151: does this need to be a string? can't we pass the path directly? i know there's no *real* difference between the 2 but it feels better to be passing a path than a string

pkgs/default.nix
```diff
@@ -116,4 +116,63 @@ rec {
...
+      };
+    });
+    lib = pkgs.lib;
+    options = eval.options;
+    filteredOptions = builtins.removeAttrs options ["_module" "description" "displayVersion" "id" "name" "replit"];
+    docsJson = (pkgs.nixosOptionsDoc {
+      options = filteredOptions;
+    }).optionsJSON;
+    in filteredOptions;
```
cdmistman, line 176: can you `nix run nixpkgs#alejandra -- <this file path>` please?

cdmistman (commented):
i'm a bit concerned about this breaking v1 nixmodule dev, are there any breaking changes here? maybe it makes sense to make a new repo?

pkgs/default.nix
airportyh: What's the suggested change? I can't see the diff...

airportyh:
> i'm a bit concerned about this breaking v1 nixmodule dev, are there any breaking changes here? maybe it makes sense to make a new repo?

We can make it backwards compatible later. "Nixmodules V2 merged to main" is [milestone 2.](https://linear.app/replit/project/nix-modules-v2-5bed040f363e)

pkgs/default.nix
airportyh: I'll do formatting in a separate pr.

pkgs/default.nix
airportyh: Okay. I also changed it to the modules directory. We are not using it anywhere though.

airportyh:
Merging this to get the prototype going. But will address concerns separately.

PR #290, opened 2024-03-19
Ruby modules

airportyh:
Why
===

Migrated Ruby modules to V2 scheme.

What changed
============

1. Migrated to v2 modules related to the Ruby programming language.
2. Added example top-level config file in `pkgs/v2/ruby.nix`. You can experiment with changing this file.

Test plan
=========

1. `nix build .#v2.ruby`
2. `cat result |jq`
3. Change `pkgs/v2/ruby.nix` and redo 1 and 2.


blast-hardcheese (approved):
Also looks sensible

PR #293, opened 2024-03-19
Node.js modules

airportyh:
Why
===

Migrate Node.js modules to V2 scheme.

What changed
============

1. Migrated Node.js module to v2 scheme.
2. Added `pkgs/v2/nodejs.nix` as an example.

Test plan
=========

1. `nix build .#v2.nodejs`
2. `cat result |jq`
3. Play with settings in `pkgs/v2/nodejs.nix` and redo 1 and 2 if desired.


blast-hardcheese (approved):
This looks sensible.

PR #297, opened 2024-03-27
Module config and registry

airportyh:
Why
===

These changes to V2 supports the prototype in https://github.com/replit/goval/pull/12896.

We now have a way to build a top-level module config that gives:
1. a `module.json` in the original module format containing toolchain that is integrated into the Replit workspace
2. a `moduleConfig.json` that gives the computed config options used by a top-level module. This can be used by the Repl configuration UI to show what tools are currently available.

We also have a registry that shows all available modules that can be installed.

What changed
============

1. added `pkg/v2/default.nix` that implements most of the logic, containing:
    * example configs and build targets
    * registry build target
    * buildDotReplit function that can build a .replit file directly
2. added descriptions to all modules
3. changed the type of env to `type = types.attrsOf types.envVar;` so that merging of `PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH` etc works
4. Temporarily commented out some flake inputs to speed up the initial build. I think due to java-language-server and fenix, we were pulling in 3 different versions of nixpkgs
5. Pinned the rev of nixpkgs-unstable to be the same as replit-nixpkgs so that it matches the artifacts on cacache

Test plan
=========

1. try building the examples like `nix build .#v2.examples.go` and view the result
2. build the registry: `nix build .#v2.registry`
3. build a .replit file:
```
nix build --expr "let flake = builtins.getFlake \"github:replit/nixmodules/th-module-config-and-registry\"; in flake.packages.x86_64-linux.v2.buildDotReplit $REPL_HOME/.replit" --impure --no-update-lock-file
```


pkgs/v2/default.nix
```diff
@@ -0,0 +1,252 @@
...
+      then
+        {
+          booleanValue = option.value;
+        }
+      else if type == "enum" || type == "str"
+      then
+        {
+          stringValue = option.value;
+        }
+      else if type == "listOf"
```
ryantm, line 145: Is this going to blow up if someone uses `types.listOf types.int`? Probably okay for now but maybe we should flag with a comment.

ryantm (approved):

pkgs/v2/default.nix
airportyh: Ok I can make it fail.

airportyh: Actually skip emit if it list of something else.

PR #300, opened 2024-04-01
bun 1.1.0

cdmistman:
Why
===

*plays Two And A Half Men theme song but it's "bun one one none bun v one one none"*

What changed
============

bun 1.1.0

Test plan
=========

`bun --version` with bun-1.1 module prints `1.1.0`
same with node modules

Rollout
=======


blast-hardcheese (approved):
Will this fix assertions in `ts-interview` being off-by-one line? (assertions reported for line 98 actually happen on 97, as an example)

PR #305, opened 2024-04-04
Queryable bundles

airportyh:
Why
===

This implements queryable bundles described in [this proposal](https://docs.google.com/document/d/1NqiSSCfRXlWgNiZpZksWqOnWg0Lb5w7Ra8YRr7wtFvA).

What changed
============

1. created mkBundleModule helper for creating bundles that expose their submodules as metadata.
2. rewrote existing bundles to use it

Test plan
=========

1. `nix build .#v2.registry && cat result |jq`
2. see that `bundles.bun`, `bundles.go`, `bundles.ruby`, `bundles.nodejs`, and `bundles.web` have a `submodules` field with a default value that contains the IDs of its submodules.
3. `nix build .#v2.examples.web`
4. `cat result/modules.json | jq` see that it contains the language servers it should have based on `./pkgs/v2/examples/web.nix`


pkgs/bundle-util/default.nix
```diff
@@ -0,0 +1,98 @@
...
+      config - an existing nested config, e.g.
+        {
+          interpreters = {
+            nodejs = {
+              enable = mkDefault true;
+            };
+          };
+        }
+      */
+      addEnableConfig = elems: config:
```
cdmistman, line 46:
```suggestion
      enablePath = elems: config:
```

cdmistman: in general i think this function would be better implemented via https://noogle.dev/f/lib/attrsets/zip to ensure that the behavior is consistent with other modules' behavior

cdmistman: i'd much rather see this function implemented in terms of https://noogle.dev/f/lib/setAttrByPath tbh

pkgs/bundle-util/default.nix
```diff
@@ -0,0 +1,98 @@
...
+        if elems == [ ]
+        then
+          { enable = mkDefault true; }
```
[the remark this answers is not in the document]
airportyh: where wouldn't be anything to merge here.

pkgs/bundle-util/default.nix
airportyh: After calling setAttrByPath you'd still need to traverse the graph a second pass to merge it with the other config. Doable but I'd have to write another function to do such merge. It's a tradeoff. I don't see a problem with the current implementation.

cdmistman (approved):

PR #309, opened 2024-04-10
introducing: per-module checks

cdmistman:
Why
===

we want the testing story to be a lot better for v2 than it was for v1

this testing system allows us to write completely hermetic checks inline with the rest of the module declarations, while also allowing us to test various combinations of modules

the system is *very* bare-bones but i expect it to grow as we define needs

What changed
============

- added new `check` function as an output to `v2/default.nix`
- added new `checks` option 

Test plan
=========

for now, can be tested like so:
```sh
$ nix repl
> :lf github:replit/nixmodules/cad/module-checks
> :b packages.x86_64-linux.v2.check { interpreters.nodejs.enable = true; }
  out -> /nix/store/abc123-modules-checks
$ /nix/store/abc123-modules-checks
```

image

note that passing in 0 enabled modules will result in an empty checks file

Rollout
=======

### DIFF
--- a/pkgs/modules/interpreters/nodejs/default.nix
+++ b/pkgs/modules/interpreters/nodejs/default.nix
@@ -36,6 +36,14 @@ with pkgs.lib; {
   };
 
   config = mkIf cfg.enable {
+    checks.nodejs = pkgs.writeShellScript "nodejs-check" ''
+      output=$(${nodejs-wrapped}/bin/node --version)
+      if [ "$output" != "v${nodejs-wrapped.version}" ]; then
+        echo "Node.js version mismatch: expected v${nodejs-wrapped.version}, got $output"
+        exit 1
+      fi
+    '';
+
     replit = {
       packages = [
         nodejs-wrapped
--- a/pkgs/moduleit/module-definition.nix
+++ b/pkgs/moduleit/module-definition.nix
@@ -637,6 +637,12 @@ in
       };
 
     };
+
+    checks = mkOption {
+      type = types.attrsOf types.package;
+      default = { };
+      description = "module tests";
+    };
   };
 
   config = {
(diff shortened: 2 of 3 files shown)


pkgs/modules/interpreters/nodejs/default.nix
```diff
@@ -36,6 +36,14 @@ with pkgs.lib; {
...
 
   config = mkIf cfg.enable {
+    checks.nodejs = pkgs.writeShellScript "nodejs-check" ''
+      output=$(${nodejs-wrapped}/bin/node --version)
+      if [ "$output" != "v${short-version}" ]; then
+        echo "Node.js version mismatch: expected v${short-version}, got $output"
+        exit 1
+      fi
+    '';
+
```
cdmistman, line 46: this is something that would've caught a bug in a previous pid1 rollout :D

cdmistman: (left this comment before pushing the more recent version of the test, oops)

pkgs/modules/interpreters/nodejs/default.nix
```diff
@@ -36,6 +36,14 @@ with pkgs.lib; {
   };
 
   config = mkIf cfg.enable {
+    checks.nodejs = pkgs.writeShellScript "nodejs-check" ''
+      output=$(${nodejs-wrapped}/bin/node --version)
+      if [ "$output" != "v${nodejs-wrapped.version}" ]; then
+        echo "Node.js version mismatch: expected v${nodejs-wrapped.version}, got $output"
+        exit 1
+      fi
```
cdmistman, line 44: this test could've caught a bug in a previous release of v1!

airportyh: Yes, this will be very important when we enforce version pinning. Do not want to accidentally upgrade the version by upgrading a Nix channel.

airportyh (changes requested):
Could you setup a way to run all module checks as a command that can be used in ci?

ryantm:
We can reopen if we revisit.

PR #311, opened 2024-04-11
feat/Add python3.12 module

blast-hardcheese:
Why
===

Python 3.12 exists, we should support it!

What changed
============

Python 3.12 was added!

Test plan
=========

Manual testing, including poetry (newly published against 3.12), pip, and language server

Rollout
=======


cdmistman (approved):
i like supporting things

PR #313, opened 2024-04-16
fix for string list serialization in protobuf

airportyh:
Why
===

Protobuf JSON serialization generated by Nix code was wrong. We need a nested object for string list value than wraps a string array. This broke the NixModulesChange event as well as the ModuleConfigGetRequest/Response.

What changed
============

Added the nested object.

Test plan
=========

1. have nixmodules v2 flag on
2. put `bundles.nodejs.enable = true` in `[moduleConfig]` section
3. wait for a NixModulesChange event that gives you moduleV2 field that is populated.

### DIFF
--- a/pkgs/v2/default.nix
+++ b/pkgs/v2/default.nix
@@ -159,7 +159,9 @@ let
     else if isListOfStr option
     then
       {
-        stringListValue = option.value;
+        stringListValue = {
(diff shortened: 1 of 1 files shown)


blast-hardcheese (approved):
lgtm

PR #314, opened 2024-04-17
Python module conversion to V2

airportyh:
Why
===

Have Python work in V2 Nixmodules.

What changed
============

1. broke out interpreter, lsp, debugger, packager, and a bundle to automatically enable them.
2. added a python-versions map for selecting a python version. Written to mimic what overlang looks like. This may be swapped out for overlang later.
3. extracted overlay into a file
4. packagers/python disables itself automatically if it detects python version is 3.8 (it doesn't build for unstable because sphinx is of an invalid version)

Test plan
=========

1. `nix build .#v2.examples.python`
2. `ls result`
5. Try changing the options in `pkgs/v2/examples/python.nix` and repeat.


pkgs/modules/packagers/python/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  options = {
+    packagers.python = {
+      enable = mkModuleEnableOption {
+        name = "Python packager";
+        description = "Python packaging with Poetry or Pip";
+      };
+    };
+  };
+
+  config = mkIf (cfg.enable && versionSupported) {
```
airportyh, line 40: Here we forcibly disable packager for Python 3.8 because pip doesn't build for Python 3.8 in unstable and I don't bother fixing it.

airportyh: But the rest of the python bundle still can work for 3.8!

pkgs/modules/bundles/python/default.nix
```diff
@@ -0,0 +1,13 @@
...
+pkgs.lib.mkBundleModule {
+  id = "python";
+  name = "Python Tools Bundle";
+  description = "Development tools for the Python 3 runtime";
+  submodules = [
+    "interpreters.python"
+    "languageServers.pyright-extended"
+    "packagers.python"
+    "debuggers.debugpy"
+  ];
```
cdmistman, line 11: hm, so overlang has `.` in some of its attrs. for example:
```
pkgs.nodejsVersions."20.2.0"
```

i'm wondering if there's a scenario where these patterns will conflict? not something that needs to be addressed now, just a thought

pkgs/modules/interpreters/python/default.nix
```diff
@@ -0,0 +1,57 @@
...
+
+      version = mkOption {
+        type = types.enum (attrNames pkgs.python-versions);
+        description = "Python version";
+      };
+
+      useSiteCustomize = mkOption {
+        type = types.bool;
+        description = "Whether to use Replit's sitecustomize.py";
+        default = true;
```
cdmistman, line 33: since we're targeting compat with other systems, it'd be nice to eventually be able to do something like `default = cfg.isReplit`, again just a thought

pkgs/modules/bundles/python/default.nix
airportyh: Why did you bring it up wrt to submodules? The python bundle tries to not be concerned with version and let that be set at the python interpreter level.

pkgs/modules/interpreters/python/default.nix
airportyh: Interesting. I can convert the replit module, and then we'd get another example.

pkgs/modules/packagers/python/default.nix
blast-hardcheese: 3.8 is officially EOL 2024-10, so only 6 more months until more of the ecosystem stops maintaining backwards compatibility

blast-hardcheese (approved):

cdmistman (approved):
lgtm

PR #315, opened 2024-04-17
Convert replit module to V2

airportyh:
Why
===

Convert replit module to V2 scheme.

What changed
============

1. Moved dotreplit-lsp over
2. Added a replit bundle
3. Set the default enable value for replit bundle to true. This gives users the ability to opt-out by setting `bundles.replit.enable = false`

Test
===

1. `nix build .#v2.examples.replit`
2. `ls result`

### DIFF
--- a/pkgs/modules/languageServers/dotreplit-lsp/default.nix
+++ b/pkgs/modules/languageServers/dotreplit-lsp/default.nix
@@ -1,0 +1,59 @@
+{ pkgs, config, ... }:
+with pkgs.lib;
+let
+  cfg = config.languageServers.dotreplit-lsp;
+  # use changes from https://github.com/tamasfe/taplo/pull/510
+  # the above PR adds the `-c` flag to the `taplo lsp` command, which is
+  # necessary to support our schema. There are a couple of paths forward
+  # for replacing this derivation:
+  # option 1:
+  # - the above pr is merged
+  # - new taplo version is released https://github.com/tamasfe/taplo/pull/502 (with my change)
+  # - nixpkgs-unstable gets the updated version
+  # - we update to nixpkgs-unstable that contains the updated taplo version
+  # option 2:
+  # - given that taplo currently consumes >1 mb of memory, it'd be nice to have
+  #   a custom bin that wraps taplo-lsp crate that *only* provides lsp for .replit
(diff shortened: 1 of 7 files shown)


pkgs/modules/languageServers/dotreplit-lsp/default.nix
```diff
@@ -0,0 +1,59 @@
+{ pkgs, config, ... }:
+with pkgs.lib;
+let
+  cfg = config.languageServers.dotreplit-lsp;
+  # use changes from https://github.com/tamasfe/taplo/pull/510
```
airportyh, line 5: btw @cdmistman your stuff's been merged.

airportyh:
Re: https://github.com/replit/nixmodules/pull/314#discussion_r1568988605 it's possible to default Python's `useSiteCustomize` to true when replit bundle is enabled, but I am not sure that's what we want, because a user wanting to disable the replit bundle doesn't mean they are not running in a repl.

pkgs/modules/languageServers/dotreplit-lsp/default.nix
cdmistman: yup! was able to get it merged after pinging. still hasn't been a release yet though so once that's done we can go back to relying on their standard releases :)

cdmistman:
> a user wanting to disable the replit bundle doesn't mean they are not running in a repl.

agreed, looks like we'll want some stdenv-style magic attrs that represent the nature of the user's environment

pkgs/modules/languageServers/dotreplit-lsp/default.nix
```diff
@@ -0,0 +1,59 @@
...
+  # - use the above custom bin
+  taplo = pkgs.rustPlatform.buildRustPackage rec {
+    pname = "taplo";
+    version = "0.patched";
+    src = pkgs.fetchFromGitHub {
+      owner = "cdmistman";
+      repo = "taplo";
+      rev = "22eff1f7775e48eee8b50518c67f992b4595ab61";
+      hash = "sha256-63fm8pH03TJd4QBuhIxtttoEAaBnc9TuHGKCMK4YGP0=";
+    };
```
cdmistman, line 27: i'll update this soon to use their latest release + my patch, no action needed here though

cdmistman (approved):

PR #319, opened 2024-04-19
Create convert_100doc.sh

ryantm:
Why
===

we need a script to covert 100 days of code projects to use nix modules

What changed
============

add the script

Test plan
=========

try it when converting one and make sure it works

Rollout
=======


blast-hardcheese (approved):
Quite handy, thank you.

PR #320, opened 2024-04-19
Submodules rename to dependencies

airportyh:
Why
===

I realized that we may use dependencies between modules more than I thought we were going to use it. For example, if you want to add the upm Node.js packager, you'll need to install Node.js. So really the packager depends on the runtime. It seems wrong to say the runtime in a submodule of the packager. So let's rename it.

What changed
============

Renamed submodule to dependencies in all the places.

Test plan
=========

1. `nix build .#v2.registry && cat result |jq|less`
2. See that `bundles.bun` and any bundle for that matter, contains a dependencies option, containing a list of module IDs

### DIFF
--- a/pkgs/bundle-util/default.nix
+++ b/pkgs/bundle-util/default.nix
@@ -2,8 +2,8 @@
   /*
     This utility creates a module that implements a bundle, which has
     these features:
-    1. if the bundle is enabled, it enables by default a list of submodules
-    2. exposes its submodules as a list of module ID strings via the `submodules`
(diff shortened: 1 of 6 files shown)


airportyh:
On second thought, dependencies doesn't seem right for a bundle and its constituents...

cdmistman (commented):
> dependencies doesn't seem right for a bundle and its constituents

agreed

airportyh:
Closing due to de-prioritizing V2..

PR #323, opened 2024-04-30
Replit RTLD loader module

airportyh:
Why
===

We need a way to deliver RTLD loader globally to all repls, but the previous attempt had issues in deployment. The new strategy is to deliver it via a Nixmodule which is conditionally turned on via a flag.

What changed
============

1. Added a module for delivering the RTLD loader
2. Added a upgrade map entry for it so we can activate it using a bare module name

Test plan
=========

See test plan in https://github.com/replit/goval/pull/13414


blast-hardcheese (approved):
looks straightforward enough

PR #324, opened 2024-05-02
bun 1.1.6

cdmistman:
Why
===

bun go brrrrr

What changed
============

- updated bun to 1.1.6
- also found out that the bun-1.1 module hasn't been getting upgraded :sad_mario:

Test plan
=========

`bun --version` prints 1.1.6

Rollout
=======


lafkpages:
In a new Bun Repl, `bun -v` still prints `1.1.0`. How long does it take for changes in this repo to be reflected in production?

lafkpages:
Tried it again just now and it's now `1.1.6`, even though the `modules` entry in `.replit` was still set to the old `bun-1.1:v1...`. Seems like I answered my own question.

PR #325, opened 2024-05-02
remove tsserver override

cdmistman:
Why
===

the `src` for typescript-language-server was overridden because the version of nixpkgs-unstable we were using didn't have a patch we needed

we've since updated nixpkgs-unstable quite a few revisions and now we're far past the requirement for removing this override, so now it's unoverridden

What changed
============

removed `src` override of typescript-language-server

Test plan
=========

1. open node 18 repl, lsp still works
2. open node 20 repl, lsp still works
3. open bun repl, lsp still works
4. open typescript repl, lsp still works

Rollout
=======

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -511,6 +511,10 @@
     "commit": "5b9002a74d16fa9d462dbc9ada6b8239ad06bed6",
     "path": "/nix/store/l5agplpqf0zijrr128kcnnsiq6mqgkr6-replit-module-nodejs-18"
   },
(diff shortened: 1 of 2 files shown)


ryantm (approved):
Nice. Next time we could add a nix assertion on the version which would have let us know that we could remove this when the we had the right version.

cdmistman:
ohhhhh good point! will do that next time :)

airportyh (approved):

PR #326, opened 2024-05-02
update fenix

cdmistman:
Why
===

there have been some rust releases since the last time fenix was updated

What changed
============

updated fenix input and locked modules

Test plan
=========

rust-stable and rust-nightly modules still work as intended

Rollout
=======


blast-hardcheese (approved):
Yee

PR #327, opened 2024-05-06
Remove modules.json lock file

airportyh:
Why
===

We want to remove `module.json` lock from the Nixmodules repo and the need to update that file. This will

* remove dev friction
* simplify the system by removing a complex version system that doesn't bring enough value

Related to:
* https://github.com/replit/goval/pull/13482
* https://github.com/replit/goval/pull/13485
* https://github.com/replit/goval/pull/13486

What changed
============

1. Removed files associated with the upgrade system
2. Replaced auto upgrade system with a simpler one in `pkgs/upgrade-map`
3. Added `pkgs/historical-modules` which allows us to still provide modules from older versions of this repo to the disk
4. Added `pkgs/bundle` which creates a linkFarm derivation containing `etc/nixmodules` directory and symlinks to module JSON files
5. Changed CI script (an addition to build any changed modules from upstream will be added separately https://github.com/replit/nixmodules/pull/328)

Test plan
=========

1. `nix build .#registry` works (but will be slow)
2. `nix build .#activeRegistry` works and is faster
3. `nix eval .#activeModules --json` lists all active modules
4. `nix eval .#modules --json` but will be slow
6. `nix build .#custom-bundle` works
7. `nix build .#bundle-squashfs` works
8. `nix build .#bundle-image-tarball` works (this should be tested in ci)


pkgs/default.nix
```diff
@@ -6,115 +6,66 @@ let
...
+    modules = self.activeModules;
+  };
 
-  rev = pkgs.writeText "rev" revstring;
+  registry = pkgs.callPackage ./registry {
+    inherit modules;
+  };
 
-  rev_long = pkgs.writeText "rev_long" revstring_long;
+  bundle = bundle-fn { };
```
ryantm, line 35: thank you for simplifying this, it was bugging me

pkgs/default.nix
```diff
@@ -6,115 +6,66 @@ let
...
 
-  dev-module-ids = [ "python-3.10" "nodejs-18" "nodejs-20" "docker" "replit" "replit-rtld-loader" ];
-
+  # For dev use: builds the shared Nixmodules disk
   bundle-squashfs = bundle-squashfs-fn {
     moduleIds = dev-module-ids;
     diskName = "disk.sqsh";
   };
 
+  # For dev use: with goval's `make custom-nixmodules-disk`
```
ryantm, line 56: I wonder if we don't really need this anymore since you can develop nix modules in Repls.

ryantm (approved):

pkgs/default.nix
airportyh: Still need the disk setup to deploy the registry.json and modules.json and just e2e setup.

airportyh:
Will wait to merge until https://github.com/replit/goval/pull/13485 is deployed.

PR #328, opened 2024-05-07
Build changed modules in ci

airportyh:
Why
===

After https://github.com/replit/nixmodules/pull/327 we need to update the CI check for building changed modules.

What changed
============

1. Updated build_changed_modules.py script to base the difference on the tuple `(module Id, sha from Nix store path)`
2. also added a check to make sure we didn't remove any module IDs (we could force upgrade them if we needed, but not out right remove)

Test plan
=========

1. Try making a subbranch of this, make a change to a module and push, see that the changed module is built in CI
2. Try removing a module from `pkgs/modules/default.nix` or `pkgs/historical-modules/default.nix` and see that CI fails

### DIFF
--- a/scripts/build_changed_modules.py
+++ b/scripts/build_changed_modules.py
@@ -1,19 +1,29 @@
-# This script test builds the updated modules for a pull request branch.
+# This script test builds the updated modules against an upstream branch for a pull request.
 
 import json
 import os
 import subprocess
 import argparse
+import re
 
-module_registry_file = 'modules.json'
+nix_store_path_pattern = re.compile(r'/nix/store/([0-9a-z]+)')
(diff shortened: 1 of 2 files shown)


blast-hardcheese (approved):
When does this get run? I'd hope that this gets run on main itself, instead of having alerts for main show up in other places

airportyh:
> When does this get run? I'd hope that this gets run on main itself, instead of having alerts for main show up in other places

It runs on the feature branch against main. If it ran on main it'd be a noop though, because it does a diff between the module set of the current branch vs main.

airportyh:
Issues in main shouldn't show up in the feature branch.

PR #330, opened 2024-05-13
Removed the duplicate go: both resulted to the same output path

airportyh:
Why
===

Noticed we have 2 go modules in the list, but only one in the output with:

```
nix eval .#activeModules --json
```

Regardless of which I comment out, the output path is the same.

What changed
============

Removed the extra one.

Test plan
=========

Go should still work.

### DIFF
--- a/pkgs/modules/default.nix
+++ b/pkgs/modules/default.nix
@@ -43,7 +43,10 @@ let
     (import ./nodejs-with-prybar)
 
     (import ./go {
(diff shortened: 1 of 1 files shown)


ryantm (commented):
I'd prefer to keep the one you removed or somehow otherwise ensure that gopls is using the same version of go. Connor told me that sometimes Nixpkgs gets out of sync.

ryantm (approved):

PR #331, opened 2024-05-13
Updated nixpkgs-stable, nixd; added docker-compose module

airportyh:
Why
===

We need the latest version of docker compose.

What changed
============

1. updated nixpkgs-unstable
3. updated nixd flake as well because it broke after the nixpkgs-unstable update

Test plan
=========

1. create a repl
2. add "docker" to `modules` array in `.replit`
3. try to use docker-compose in shell and see its version is 2.27.0.

### DIFF
--- a/pkgs/historical-modules/default.nix
+++ b/pkgs/historical-modules/default.nix
@@ -44,6 +44,14 @@ let
       commit = "4c6f2315da24b84bd5e9dfedb952e41677724aaa";
     }
     {
+      moduleId = "haskell-ghc9.4";
+      commit = "4d6f7cd9fd685a3319ac7b6e3fc0789b430d6289";
+    }
+    {
+      moduleId = "elixir-1_15";
+      commit = "4d6f7cd9fd685a3319ac7b6e3fc0789b430d6289";
+    }
+    {
       moduleId = "nodejs-14";
       commit = "f4cd419a646009297c049a2f1eec434381e08f13";
     }
--- a/pkgs/modules/python/pip.nix
+++ b/pkgs/modules/python/pip.nix
@@ -1,7 +1,16 @@
 pkgs @ { pypkgs, ... }:
 
 let
-  pip = pypkgs.pip;
+  pip = pypkgs.pip.overridePythonAttrs (old: rec {
+    outputs = [
+      "out"
+    ];
+    # Skip building the docs for pip because that was failing
+    # with a sphinx-build error with a version of nixpkgs-unstable for 3.10 (worked for >3.11)
+    # and we don't need the docs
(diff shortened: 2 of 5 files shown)


pkgs/modules/default.nix
```diff
@@ -60,6 +60,7 @@ let
     (import ./dart)
     (import ./deno)
     (import ./docker)
+    (import ./docker-compose)
```
ryantm, line 63: https://github.com/replit/nixmodules/blob/4d6f7cd9fd685a3319ac7b6e3fc0789b430d6289/pkgs/modules/docker/default.nix#L182 docker-compose is already part of the docker module. I think we don't need to add another module for this.

pkgs/historical-modules/default.nix
```diff
@@ -43,6 +43,14 @@ let
       moduleId = "haskell-ghc9.2";
       commit = "4c6f2315da24b84bd5e9dfedb952e41677724aaa";
     }
+    {
+      moduleId = "haskell-ghc9.4";
```
airportyh, line 47: latest is haskell-ghc9.6

pkgs/historical-modules/default.nix
```diff
@@ -43,6 +43,14 @@ let
...
+      commit = "4d6f7cd9fd685a3319ac7b6e3fc0789b430d6289";
+    }
+    {
+      moduleId = "elixir-1_15";
```
airportyh, line 51: latest is elixir-1_16

ryantm (approved):

Juanito98:
🚢

pkgs/modules/python/pip.nix
```diff
@@ -1,7 +1,16 @@
...
-  pip = pypkgs.pip;
+  pip = pypkgs.pip.overridePythonAttrs (old: rec {
+    outputs = [
+      "out"
+    ];
+    # Skip building the docs for pip because that was failing
+    # with a sphinx-build error with a version of nixpkgs-unstable for 3.10 (worked for >3.11)
+    # and we don't need the docs
+    postBuild = "";
+    postInstall = "";
```
ryantm, line 12: Are we also going to lose [shell completions here](https://github.com/NixOS/nixpkgs/blob/9ddcaffecdf098822d944d4147dd8da30b4e6843/pkgs/development/python-modules/pip/default.nix#L96)? It looks like they aren't working anyway already though.

ryantm (approved):

pkgs/modules/python/pip.nix
airportyh: Oops I can add them back in another pr (ci will be quicker).

PR #339, opened 2024-06-03
bun: update to 1.1.11

cdmistman:
Why
===

bun 1.1.11 almost rhymes if you squint hard enough

What changed
============

bun to 1.1.11

Test plan
=========

`bun --version` prints the right version

Rollout
=======

### DIFF
--- a/pkgs/bun/default.nix
+++ b/pkgs/bun/default.nix
@@ -3,9 +3,9 @@
 }:
 
 bun.overrideAttrs rec {
-  version = "1.1.8";
+  version = "1.1.11";
   src = fetchurl {
     url = "https://github.com/oven-sh/bun/releases/download/bun-v${version}/bun-linux-x64.zip";
-    hash = "sha256-JuDT+8FHbamdMVCDeSTGPYOvhPY2EZ9XeD2zjHEj36Y=";
+    hash = "sha256-RfP4fnTqumZKMBvC3ze4Lb7cQuUVl/czL+8xqB00kPA=";
   };
 }


ryantm (approved):

lafkpages:
For the version hashes, check out the NixOS docs on [Nix Hashes](https://nixos.wiki/wiki/Nix_Hash). It's the same as the `integrity` attributes for HTML `<script>` and `<link>` tags. For example, if you download Bun v1.1.8 from the URL in the Nix module edited in this PR, and run `openssl dgst -sha256 -binary bun-linux-x64.zip | openssl base64 -A`, it gives you the same hash as in the deleted changes of this PR.

cdmistman:
hey @7heMech

> I get version 1.1.6

`bun-1.1` will give you the latest available `1.1.x` version of bun available on replit. so since you're getting 1.1.6, that means that i forgot to update the disk after https://github.com/replit/nixmodules/commit/c52d26f69bd0a1d395070854f645b696cfe883bc was merged

i did kick off a disk build after this PR landed, but it takes a few hours (still not done, ~2 hours later), and can take up to a couple more hours after it's finished building to attach to your Repl, so please have patience

> older strings don't automatically update

correct, that's intentional. we want to upgrade users to newer version-compatible versions when possible, so the upgrade mechanism is based on the module ID. as of a couple of weeks ago, the module ID is solely `bun-1.1` for this module, for example - it used to be something like `bun-1.1:abc123`. that mechanism is no longer in use, but is still compatible with the current mechanism. as such, repls with `bun-1.1:abc123` format will get bun 1.1.11 when it's released, as well as any repls using `bun-1.1`.

> Oh, and how do you make the version hashes?

luis's suggestion works but is a bit harder for me to remember. the easiest method:
1. remove the hash string
2. `nix build '.#"bun-1.1"'` (note that the quoting is necessary here)
6. ^ prints the hash string, put it in the file

note that this also works if you replace steps 2-5 with `nix build '.#"bun-1.1"'`, i use `nix repl` as a point of preference

PR #340, opened 2024-06-04
Supply historical displayVersions

blast-hardcheese:
Why
===

Supply `displayVersions` for historical modules

What changed
============

These modules don't have `displayVersions`. Hardcode them.

NB: Not all module id versions line up with what they actually ship:
image

Test plan
=========

N/A, this is just informative.

Rollout
=======

### DIFF
--- a/pkgs/historical-modules/default.nix
+++ b/pkgs/historical-modules/default.nix
@@ -10,78 +10,97 @@ let
     {
       moduleId = "bun-1.0";
       commit = "fdbd39619614e07d1ff44cfe20de2ad56eb8a40f";
+      displayVersion = "1.0.23";
     }
     {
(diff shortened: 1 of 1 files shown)


ryantm (approved):
> NB: Not all module id versions line up with what they actually ship

Oof. Thanks for going through these!

airportyh:
> NB: Not all module id versions line up with what they actually ship:
that's due to the upgrade map: https://github.com/replit/nixmodules/blob/main/pkgs/upgrade-map/default.nix#L13

blast-hardcheese:
Closing this since it's included in the next PR

blast-hardcheese:
I didn't realize we were stacking PRs. I'm here for it.

PR #342, opened 2024-06-05
Added internal and demoted flags

airportyh:
Why
===

1. We'd like to hide the docker module from general users because it won't work for them
2. We'd like to show the prybar modules less prominently than the main python and Node.js modules

What changed
============

1. added an internal flag and used it for docker
2. added a demoted flag and used it for prybar modules

Test plan
=========

1. `nix build .#docker && cat result |jq` and see that its internal field is true
2. `nix build .#'"python-with-prybar-3.10"' && cat result |jq` and see it's demoted field is true

### DIFF
--- a/pkgs/moduleit/module-definition.nix
+++ b/pkgs/moduleit/module-definition.nix
@@ -622,6 +622,23 @@ in
       default = "";
     };
 
+    internal = mkOption {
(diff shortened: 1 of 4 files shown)


ryantm (approved):
Looks good. Names make sense to me.

PR #344, opened 2024-06-05
use jq for historical displayVersion

ryantm:


pkgs/historical-modules/default.nix
```diff
@@ -113,16 +113,14 @@ let
...
-      moduleJSON = fromJSON (builtins.unsafeDiscardStringContext (readFile module.outPath)) // {
-        inherit displayVersion;
-      };
     in
-    pkgs.writeText "replit-module-${moduleId}" (toJSON moduleJSON);
+    pkgs.stdenvNoCC.mkDerivation {
+      name = "replit-module-${moduleId}";
+      buildCommand = ''
+        set -x
+        ${pkgs.jq}/bin/jq '.displayVersion = "${displayVersion}"' < ${module} > $out
```
blast-hardcheese, line 121:
```suggestion
        ${pkgs.jq}/bin/jq --arg displayVersion "${displayVersion}" '.displayVersion = ''$displayVersion' < ${module} > $out
```

if we're doing this anyway. iiuc `''$` should translate to a `$` in the script that gets run. If that's the case, then the suggestion here will work.

ryantm: I'm relying on interpolating in displayVersion from the Nix string there, not escaping it.  We could add displayVersion from Nix into the environment of the mkDerivation builder if you wanted to do it your way, but this seems simpler.

blast-hardcheese: Yeah, I'm just trying to handle the case where `${displayVersion}` contains something unexpected that results in not-json being emitted. Unlikely, but this is why `--arg` exists, so we can be certain. It's such a small number of entries though so it should be fine.

ryantm: Oh I see now cool!

pkgs/historical-modules/default.nix
```diff
@@ -113,16 +113,14 @@ let
...
-      moduleJSON = fromJSON (builtins.unsafeDiscardStringContext (readFile module.outPath)) // {
-        inherit displayVersion;
-      };
     in
-    pkgs.writeText "replit-module-${moduleId}" (toJSON moduleJSON);
+    pkgs.stdenvNoCC.mkDerivation {
+      name = "replit-module-${moduleId}";
+      buildCommand = ''
+        set -x
+        ${pkgs.jq}/bin/jq '."displayVersion" = "${displayVersion}"' < ${module} > $out
```
blast-hardcheese, line 121:
```suggestion
        ${pkgs.jq}/bin/jq --arg displayVersion "${displayVersion}" '.displayVersion |= $displayVersion' < ${module} > $out
```
Tested locally:
```
$ nix build .#'"nodejs-16"' && jq < result | grep displayVersion
warning: Git tree '/app/nixmodules' is dirty
  "displayVersion": "16.18.1"
```

cdmistman (approved):
i also prefer using `jq --arg` but otherwise lgtm

don't forget pr description!

PR #348, opened 2024-06-07
Interpolate displayVersion into the right place in the tree

blast-hardcheese:
Why
===

Top-level `displayVersion` is actually the semantically-relevant minified version, `.runner.[*].displayVersion` is the full version string

What changed
============

Interpolate values from the new `override` parameter

Test plan
=========

```
$ nix build .#'"nodejs-16"'
$ jq . < result | jq '.displayVersion, .runners.nodeJS.displayVersion'
"16"
"16.18.1"
```

Rollout
=======

### DIFF
--- a/pkgs/historical-modules/default.nix
+++ b/pkgs/historical-modules/default.nix
@@ -11,106 +11,193 @@ let
     {
       moduleId = "bun-1.0";
       commit = "fdbd39619614e07d1ff44cfe20de2ad56eb8a40f";
-      displayVersion = "1.0.23";
+      overrides = {
+        # /nix/store/cvz36f39793v9l361ibfxznjjmk4jpd6-replit-module-bun-1.0
+        # .runners["package.json"].displayVersion = "1.0.23";
+        displayVersion = "1";
+      };
     }
     {
       moduleId = "dart-2.18";
       commit = "c48c43c6c698223ed3ce2abc5a2d708735a77d5b";
-      displayVersion = "2.18.0";
+      overrides = {
+        # /nix/store/05i6y0nwn8zgm6250my14xbcialzak9b-replit-module-dart-2.18
+        # .runners.dart.displayVersion = "2.18.0";
+        displayVersion = "2";
+      };
     }
     {
       moduleId = "dart-3.0";
       commit = "0b7a60667d2c29f2686211e952924a9693916a20";
-      displayVersion = "3.2.4";
+      overrides = {
+        # /nix/store/0wr2lk1h8vqzs38pfwiil6cf6s5mn54j-replit-module-dart-3.2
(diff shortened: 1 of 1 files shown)


pkgs/historical-modules/default.nix
```diff
@@ -11,106 +11,193 @@ let
...
+      };
     }
     {
       moduleId = "go-1.19";
       commit = "c48c43c6c698223ed3ce2abc5a2d708735a77d5b";
-      displayVersion = "1.19.3";
+      overrides = {
+        # /nix/store/7icmam9gmbfas0zbpm53f3ilw79kx0rj-replit-module-go-1.19
+        # .runners["go-run"].displayVersion = "1.19.3";
+        displayVersion = "1";
```
ryantm, line 69: For go, I believe the minor version matters quite a bit.

pkgs/historical-modules/default.nix
```diff
@@ -11,106 +11,193 @@ let
...
+      overrides = {
+        # /nix/store/65v1j76svgdi8bj3sxx5hfshbvdyissl-replit-module-zig-0.11
+        # .runners["zig-build-run"].displayVersion = "0.11.0";
+        displayVersion = "0";
+      };
     }
   ];
 
-  moduleFromHistory = { displayVersion, moduleId, commit, deployment ? false }:
+  moduleFromHistory = { moduleId, commit, deployment ? false, overrides }:
```
ryantm, line 200: We need to fix line 143 below too

pkgs/historical-modules/default.nix
```diff
@@ -11,106 +11,193 @@ let
...
+        displayVersion = "5";
+      };
     }
     {
       moduleId = "vue-node-18";
       commit = "3ea4bcdbdc3c5e3c09b37b07edcd61781f9695f7";
-      displayVersion = "18.18.2";
+      overrides = {
+        # /nix/store/623sn0cy54r5csab3d9fmg55di6r4dd3-replit-module-vue-node-18
+        # .runners["dev-runner"].displayVersion = "18.18.2";
```
airportyh, line 185: Why are some of these commented out?

blast-hardcheese: I spot-checked and it didn't seem as though they were used, I figured I'd just fix the ones that I actually need right now and we can add the rest of the specific versions later if it turns out they're useful. I did the legwork of correlating the values and keys with the comments, so it should be easier in the next pass.

blast-hardcheese:
I believe this is now good to go.

pkgs/historical-modules/default.nix
airportyh: ok

airportyh (approved):

ryantm (approved):

PR #350, opened 2024-06-07
[ruby] revert rexml change in upstream

ryantm:
Why
===
* [a change](https://github.com/NixOS/nixpkgs/pull/312320) to rexml broke ruby 3.2 because it needed a newer version of rexml than is provide:
needed: >3.0.9
got:/nix/store/8fg60ic5fbj9qncw1m760z08zljich6q-ruby-3.2.2/lib/ruby/gems/3.2.0/gems/strscan-3.0.5/

What changed
===
* reverse that change

Test plan
===
* solagraph works on ruby 3.2

```
[ryantm@replit1:~/p/replit/nixmodules]$ /nix/store/assiad6yc2nminnvjjarfviza49mncrh-ruby3.2-solargraph-0.50.0/bin/solargraph stdio
Solargraph is listening on stdio PID=3385015
```

Rollout
=======

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -35,7 +35,16 @@
 
       pkgs-23_05 = mkPkgs nixpkgs "x86_64-linux";
 
-      pkgs = mkPkgs nixpkgs-unstable "x86_64-linux";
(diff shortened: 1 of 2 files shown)


airportyh (approved):
Thank you!

PR #353, opened 2024-06-12
Switch nix lsp from nixd to nil

vlinkz:
Why
===

[DX-679](https://linear.app/replit/issue/DX-679/installing-the-nix-module-brings-in-nixd-which-immediately-expands-to): Currently the nix lsp, `nixd`, uses >50% of replit CPU in the background. The proposed alternative, [`nil`](https://github.com/oxalica/nil), uses barely any CPU.

Resources using `nixd`
Screenshot from 2024-06-12 15-23-57

Resources using `nil`
Screenshot from 2024-06-12 15-23-08

What changed
============

Switch nix lsp from `nixd` to `nil`.

Test plan
=========

- Tested modified nix module in a repl

Rollout
=======
It would be nice to have https://github.com/replit/repl-it-web/pull/45470 merged first, as it fixes an issue where `nil` will sporadically close due to receiving a `didSave` notification.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,8 +4,6 @@
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
   inputs.fenix.url = "github:nix-community/fenix";
   inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
-  inputs.nixd.url = "github:nix-community/nixd";
-  inputs.nixd.inputs.nixpkgs.follows = "nixpkgs";
   inputs.prybar.url = "github:replit/prybar";
   inputs.prybar.inputs.nixpkgs.follows = "nixpkgs";
   inputs.java-language-server.url = "github:replit/java-language-server";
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -4,8 +4,8 @@
   inputs.nixpkgs-unstable.url = "github:nixos/nixpkgs/nixpkgs-unstable";
   inputs.fenix.url = "github:nix-community/fenix";
   inputs.fenix.inputs.nixpkgs.follows = "nixpkgs";
-  inputs.nixd.url = "github:nix-community/nixd";
-  inputs.nixd.inputs.nixpkgs.follows = "nixpkgs";
+  inputs.nil.url = "github:oxalica/nil";
+  inputs.nil.inputs.nixpkgs.follows = "nixpkgs";
```
ryantm, line 8: Do we need to use the github nil, can we use the one in nixpkgs?

vlinkz: The one in nixpkgs is around a year old [nil 2023-08-09](https://github.com/oxalica/nil/releases/tag/2023-08-09). But it shouldn't cause any issues using the old version (I tested both already)

ryantm (approved):

PR #354, opened 2024-06-13
Use `nixpkgs-fmt` as formatter for the nix LSP

vlinkz:
Why
===

By default, `nil` does not have a formatter defined. This pr adds `nixpkgs-fmt` as the formatter used by the `nil` LSP.

What changed
============

`nil` is now configured to use `nixpkgs-fmt` for formatting `.nix` files.

Test plan
=========

- Tested modified nixmodules in a repl

Rollout
=======


ryantm (approved):
Fine with using this for now, but we should probably switch to the RFC-accepted formater as it becomes available.

PR #356, opened 2024-06-17
Add `defaultEntrypoints` field to `runnerOptions`

vlinkz:
Why
===

Right now, if no entrypoint file is defined in `.replit`, even if a language module is installed and a sensible default file to run exists, it is not run. 

What changed
============

Added `defaultEntrypoints` fields to `runnerOptions`. In the case that no entrypoint file is defined, the backend can check if one of these files exist, and if so, run it instead of failing.  

Test plan
=========

- Tested modified nixmodules in a repl

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -121,6 +121,7 @@ in
     fileParam = true;
     language = "python3";
     start = "${python3-wrapper}/bin/python3 $file";
+    defaultEntrypoints = [ "main.py" "app.py" "run.py" ];
   };
 
   replit.dev.debuggers = debuggerConfig;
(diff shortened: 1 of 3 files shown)


blast-hardcheese (approved):
👍

pkgs/modules/python/default.nix
```diff
@@ -121,6 +121,7 @@ in
     fileParam = true;
     language = "python3";
     start = "${python3-wrapper}/bin/python3 $file";
+    defaultEntrypoints = [ "main.py" "run.py" ];
   };
```
ryantm, line 125: Can we add app.py to this too?

ryantm (approved):

ryantm (approved):

PR #360, opened 2024-06-25
Introduce unified formatter interface; Use for `gofmt` and `prettier`

bradymadden97:
Why
===

As I was starting to write the pid2 formatter, I was realizing that I was going to have to duplicate some of the same hacks for prettier that I added to the frontend for pid1 (which formats via exec).

We have a unified interface on the frontend to initiate formatting, but individual formatters do not have a unified set of arguments. So in order to support some special arguments for prettier (like range formatting, and tab and spaces settings) I had to explicitly pass in arguments https://github.com/replit/repl-it-web/blob/b2f9ae504907d097e137c152eb295dd7cdaba78c/client/codemirror/formatter/useFormattingProviders.ts#L30

I was going to have to do the same thing in the pid2 formatter, unless I came up with a better solution. In the past I wrote a doc proposing [writing all formatters as language servers](https://docs.google.com/document/d/1P2cY82PskewlPwHdrQaS9UYs6oCDFoLcnRrPHi9D6PE/edit#heading=h.ypv7jza8mj4o) to solve this problem. However, that was likely going to be a lot of work. We can end up in the same place, with a unified interface for the frontend to call, by just writing some bash scripts.

Here's the way it works:
1. Define a common "formatter" interface and set of options that the frontend can send. For this I used the following:
  - `-f | --file`: filepath - required
  - `-a`: if the formatter should write the changes directly, rather than returning the formatted string via stdout
  - `--range-start`: an offset, for supporting range formatting
  - `--range-end`: see above
  - the various LSP [formatting options as defined in the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions)
2. The frontend will pass its desired options according to the common spec
3. Each formatter, within nixmodules, will define its own script that will receive the parsed common options, and translate them to binary-specific options before calling its formatter binary. If it doesn't support one of the common options, it can just ignore it

_Describe what prompted you to make this change, link relevant resources_

What changed
============

I've implemented the above for both formatters we have in nixmodules today, gofmt and prettier. Both should be implemented in a backwards-compatible way to work with the current frontend setup. After this rolls out theres some frontend cleanup that can be done, that will then allow me to remove some of the TODOs in this PR.

Test plan
=========

I built both nodejs-20 and go locally using 
```
nix build .#custom-bundle
```

I then inspected the active modules using 
```
nix eval .#activeModules --json | jq
```

and manually ran the built formatter scripts against some typescript and go files to confirm they worked.

Tested e2e with frontend here https://replit.com/@replit-dev/brady-nixmodules#index.ts
```
nix build .#'"nodejs-20"'
ln -s result result.json
```
Add `result.json` to modules array in .replit
Able to format a typescript file with prettier 

```
nix build .#'"go-1.21"'
ln -s result result.json
```
Able to format a golang file with gofmt. The nice thing is this actually fixes a bug with gofmt that currently formats _all_ files, instead of just your active one.
 

Rollout
=======

### DIFF
--- a/pkgs/formatter/default.nix
+++ b/pkgs/formatter/default.nix
@@ -1,0 +1,165 @@
+{ pkgs }:
+
+pkgs.writeShellApplication {
+  name = "parse-formatter-options";
+  runtimeInputs = [ pkgs.bash ];
+  text = ''
+    #!/bin/bash
+
+    # Initialize variables
+    apply="false"
+    file=""
+    insertFinalNewline="false"
+    insertSpaces="false"
+    rangeStart=""
+    rangeEnd=""
+    stdinMode="false"
+    tabSize=""
+    trimFinalNewlines="false"
+    trimTrailingWhitespace="false"
+
+    # Function to print usage
+    print_usage() {
+      echo "Usage: $0 [-a|--apply] -f|--file <filename> [--stdin] [--range-start <offset>] [--range-end <offset>] [--tab-size <number>] [--insert-spaces] [--trim-trailing-whitespace] [--insert-final-newline] [--trim-final-newlines]"
+      echo
+      echo "Options:"
+      echo "  -h, --help                     Show this help message and exit"
+      echo "  -a, --apply                    Apply edits directly to the file (optional)"
+      echo "  -f, --file <filename>          Specify the file to be formatted (required)"
+      echo "      --stdin                    Read file input from stdin (optional)"
+      echo "      --insert-final-newline     Insert a newline at the end of the file if one does not exist (optional)"
+      echo "      --insert-spaces            Prefer spaces over tabs (optional)"
+      echo "      --range-start <offset>     Specify the character offset for formatting (optional, requires --range-end)"
+      echo "      --range-end <offset>       Specify the character offset for formatting (optional, requires --range-start)"
+      echo "      --tab-size <number>        Size of a tab in spaces (optional)"
+      echo "      --trim-final-newlines      Trim all newlines after the final newline at the end of the file (optional)"
+      echo "      --trim-trailing-whitespace Trim trailing whitespace on a line (optional)"
+
+    }
+
+    # Function to check if a value is a number
+    is_number() {
+      [[ "$1" =~ ^[0-9]+$ ]]
+    }
+
+    # Parse command-line arguments
+    while [[ "$#" -gt 0 ]]; do
+      case $1 in
+        -a|--apply)
+          apply="true"
+          shift
+          ;;
+        -f|--file)
+          file="$2"
+          shift 2
+          ;;
+        -h|--help)
+          print_usage
+          exit 0
+          ;;
+        --stdin)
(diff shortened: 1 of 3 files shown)


airportyh (approved):
Nice bash! Looks clean. Having tests for run-gofmt and run-prettier would be nice!

airportyh:
> Able to format a golang file with gofmt. The nice thing is this actually fixes a bug with gofmt that currently formats all files, instead of just your active one.
love this!

airportyh:
There may be a chicken/egg probably regarding which change to roll out first, the frontend change or the nixmodules changes.

bradymadden97:
I made this backwards compat w/ the current frontend (tested go and prettier w/ frontend).

So I think we can just:
1. roll this out w/ no frontend changes
2. then I will update the frontend to remove the prettier specific stuff and use some of the unified args
3. then i will come back here and remove some of the TODOs that were added for backwards-compat reasons

pkgs/modules/go/default.nix
```diff
@@ -1,7 +1,37 @@
...
+{ pkgs, lib, ... }:
 let
   goversion = lib.versions.majorMinor go.version;
+
+  formatter = import ../../formatter {
+    inherit pkgs;
+  };
+  run-gofmt = pkgs.writeShellApplication {
+    name = "run-gofmt";
+    runtimeInputs = [ pkgs.bash ];
```
airportyh, line 11: we need to add go as a runtime input I think.

airportyh, line 11:
```suggestion
    runtimeInputs = [ pkgs.bash go ];
```

airportyh (dismissed):
See small fix to go runtimeInputs.

airportyh (approved):

pkgs/formatter/default.nix
```diff
@@ -0,0 +1,165 @@
+{ pkgs }:
+
+pkgs.writeShellApplication {
```
ryantm, line 3: Thank you for using shellcheck.

pkgs/formatter/default.nix
```diff
@@ -0,0 +1,165 @@
...
+  name = "parse-formatter-options";
+  runtimeInputs = [ pkgs.bash ];
+  text = ''
+    #!/bin/bash
+
```
ryantm, line 8:
```suggestion
```
writeShellApplication will add this.

pkgs/modules/go/default.nix
```diff
@@ -1,7 +1,52 @@
...
+        gofmt_args+=("-w")
+      fi
+
+      # Append the file path
+      gofmt_args+=("$file")
+
+      # Execute the command
+      gofmt "''${gofmt_args[@]}"
+    '';
+    checkPhase = ''
```
ryantm, line 33: nix tests :heart_eyes:

ryantm (approved):

PR #365, opened 2024-06-27
Consistent nodejs names versions

airportyh:
Why
===

Inconsistent Node.js and related module names and display versions:

image

What changed
============

1. Removed the version number from the name for all cases, so it's not duplicated with the display version
2. Moved vue-node-18 out of historical and into active. Fewer historical modules the better, I think.

Test plan
=========

That list should look a lot cleaner.


ryantm (approved):
Thanks!

PR #366, opened 2024-06-27
[nodejs,angular] stop adding local node_modules/.bin to path

ryantm:
Why
===
* We were having issues where our local node_modules bin had a "which" binary that did not behave well.
* It isn't typical for local node_modules/.bin to be on the path, if you want something on the path, you typically use npm install --global, which is still added to the PATH after this change.

What changed
===
* Remove $REPL_HOME/node_modules/.bin from PATH

Test plan
===
* Check template tester still passes
* Check PATH does not contain that path in nodejs Repls

Rollout
=======

### DIFF
--- a/pkgs/modules/angular/default.nix
+++ b/pkgs/modules/angular/default.nix
@@ -35,7 +35,7 @@ in
     env = {
       XDG_CONFIG_HOME = "$REPL_HOME/.config";
       npm_config_prefix = "$REPL_HOME/.config/npm/node_global";
-      PATH = "$XDG_CONFIG_HOME/npm/node_global/bin:$REPL_HOME/node_modules/.bin";
+      PATH = "$XDG_CONFIG_HOME/npm/node_global/bin";
     };
 
     dev.runners.dev-runner = {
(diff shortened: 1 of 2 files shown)


airportyh (approved):
Okay, that's true. People are pretty used to npx and the like.

lafkpages:
I don't get how "This is fully backward and forward compatible" though. Existing Repls might have scripts that rely on local NPM binaries being in the PATH, and this would break them. And even if it only applies to new Repls, the same could happen for GitHub imports, right? You could try to import someone's old repo that had such a script and it would break.

I might be wrong though, I'm still trying to wrap my head around Nix.

ryantm:
Yes, you're right, this is a breaking change. I updated the checkmark.

https://docs.npmjs.com/cli/v10/configuring-npm/folders says 

> Any bin files are symlinked to ./node_modules/.bin/, so that they may be found by npm scripts when necessary.

we're pretty sure it is not conventional to put this dir on the user's PATH and we have evidence of this breaking at least one user, so we're going to move forward with it.

People can mitigate this themselves by adding an env section to their .replit if they need this:
```
[env]
PATH="$PATH:$REPL_HOME/node_modules/.bin"
```

PR #368, opened 2024-06-28
Revert "[nodejs,angular] stop adding local node_modules/.bin to path"

ryantm:
Reverts replit/nixmodules#366

We found that this breaks all Typescript forks. We'll need to have a fix for those before we reland this.


cdmistman (approved):
oofie

PR #375, opened 2024-07-15
Get python38 actually building from unstable

airportyh:
Why
===

https://github.com/replit/nixmodules/pull/371 actually didn't build python 3.8 from unstable. I missed something.

What changed
============

1. use override to actually create python38Full and python38Packages via overlay
2. use them to build the python-3.8 module

Test plan
=========

1. add `python-3.8` as a module in `modules`
2. `readelf -d $(python3 -c 'import sys; print(sys.executable)')`
3. you should see `glibc-2.39-52` in the `Library runpath`, not glibc-2.37...

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -84,6 +84,27 @@
             '';
           };
         };
+
+      } // rec {
+        python38 = pkgs.python310.override {
+          sourceVersion = {
+            major = "3";
+            minor = "8";
+            patch = "18";
+            suffix = "";
+          };
+          hash = "sha256-P/txzTSaMmunsvrcfn34a6V33ZxJF+UqhAGtvadAXj8=";
+        };
+        
+
+        python38Full = python38.override {
+          self = python38Full;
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -84,6 +84,27 @@
...
+            suffix = "";
+          };
+          hash = "sha256-P/txzTSaMmunsvrcfn34a6V33ZxJF+UqhAGtvadAXj8=";
+        };
+
+
+        python38Full = python38.override {
+          self = python38Full;
+          pythonAttr = "python38Full";
+          bluezSupport = true;
```
blast-hardcheese, line 103: why are we even bothering to build bluetooth support in a repl?

blast-hardcheese (approved):
question but +1 otherwise

flake.nix
airportyh: It's bluez for audio support I think, not bluetooth. I anycase I just copied from upstrteam.

PR #377, opened 2024-07-24
python: colocate sitecustomize with python derivation output

cdmistman:
Why
===

pylint makes some assumptions about the format of `PYTHONPATH` 🙄

What changed
============

changed the python wrapper to use the output of a new derivation that:
1. copies the provided `python` derivation output
2. removes its `sitecustomize.py` if it exists
3. copies in our own `sitecustomize.py`

Test plan
=========

python template tests pass

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -1,5 +1,5 @@
 { python, pypkgs }:
-{ pkgs-unstable, pkgs-23_05, lib, ... }:
+{ pkgs-unstable, lib, ... }:
 let
   pythonVersion = lib.versions.majorMinor python.version;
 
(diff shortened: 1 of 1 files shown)


pkgs/modules/python/default.nix
```diff
@@ -100,6 +99,14 @@ let
 
   sitecustomize = pkgs.callPackage ./sitecustomize.nix { };
 
+  sitecustomized-python = pkgs.runCommand "python-replit-sitecustomized" {} ''
+    ls -al ${python.outPath}
+    ${pkgs.rsync}/bin/rsync -av --chown=$USER --chmod=+w ${python.outPath}/ $out/
```
blast-hardcheese, line 104: I am baffled that this is the only way. Doesn't this double the disk required for all python packages?

cdmistman: ooooh i think you're right, i don't actually force this python's usage anywhere. i'll fix that in the morning

airportyh:
As mentioned in your one-on-one chat. I'd like to abandon making changes to python and patch the repo unless we get more data points that this is a widespread problem.

PR #378, opened 2024-07-27
Bun update

7heMech:
Why
===
In these few new versions, bun has become faster and smaller for Linux.


cdmistman (approved):
thank you!

PR #382, opened 2024-08-14
Bun update

7heMech:
Why
===
~~We're two versions behind  🗣️🗣️🗣️🔥🔥🔥 😅~~
4mb smaller, 50% faster console.log, 30% faster fetch decompression...

### DIFF
--- a/pkgs/bun/default.nix
+++ b/pkgs/bun/default.nix
@@ -3,9 +3,9 @@
 }:
 
 bun.overrideAttrs rec {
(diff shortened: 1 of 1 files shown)


ryantm (approved):

lafkpages:
@7heMech v1.1.24 is out and fixes an important regression in v1.1.23 so you might want to update to that

ryantm (approved):

PR #386, opened 2024-08-28
feat/add uv

blast-hardcheese:
Why
===

I'd like a more recent version of `uv` to address https://github.com/replit/upm/issues/244

What changed
============

- Bumping flake lock
- Disabling defunct solargraph patch for now, as it was not actually making solargraph work
- Adding `uv` to python modules
- Adding dart, elixir, and R to the upgrade map
- Deleting now-promoted `discard-references` experimental flag, as indicated in https://github.com/NixOS/nix/blob/master/doc/manual/src/release-notes/rl-2.18.md

Test plan
=========

`which uv` should show at least version 0.3.3

Rollout
=======

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -39,8 +39,15 @@
         name = "nixpkgs-unstable-patched";
         src = nixpkgs-unstable;
         patches = [
-          # rexml breaks this version of nixpkgs
-          ./patches/rexml.patch
+          # rexml broke solargraph at some point in the past.
+          # Attempting to run:
+          #   echo 'x = (5 + "hey")' > main.rb
+          #   /nix/store/.../ruby /nix/store/.../bin/solargraph typecheck main.rb
+          # curently emits a bunch of "Unrecognized RBS type:" warnings, followed by
+          # "0 problems found."
+          # This patch used to apply, but even when massaged to apply to the
+          # current unstable HEAD, still produces the same result:
+          # ./patches/rexml.patch
         ];
       };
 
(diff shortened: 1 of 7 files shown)


pkgs/modules/python/default.nix
```diff
@@ -113,6 +113,7 @@ in
     python3-wrapper
     pip-wrapper
     poetry-wrapper
+    pkgs.uv
```
ryantm, line 116: Do we need the python wrapper for uv like we're doing for pip and poetry?

blast-hardcheese: I don't believe so. I'm still doing manual testing, but `uv` relies so little on the environment there's _probably_ not much use

ryantm: great!

flake.nix
```diff
@@ -39,8 +39,15 @@
...
-          ./patches/rexml.patch
+          # rexml broke solargraph at some point in the past.
+          # Attempting to run:
+          #   echo 'x = (5 + "hey")' > main.rb
+          #   /nix/store/.../ruby /nix/store/.../bin/solargraph typecheck main.rb
+          # curently emits a bunch of "Unrecognized RBS type:" warnings, followed by
+          # "0 problems found."
+          # This patch used to apply, but even when massaged to apply to the
+          # current unstable HEAD, still produces the same result:
+          # ./patches/rexml.patch
```
ryantm, line 50: If solargraph is working, we can probably just totally remove this patch. This patch was fixing a crashing bug in solargraph

ryantm (approved):

flake.nix
blast-hardcheese: it is not crashing, but it's also not working. That's why I wrote up my findings, as to not get lost in the weeds on an incidental thing.

ryantm (approved):

blast-hardcheese:
I did some manual testing, it looks like setting the `UV_PROJECT_ENVIRONMENT` to `.pythonlibs` works, though I'll need to dust off the auto-healing functionality of the Python environ fsck to make sure we clean up stale links.

A lot of manual testing will be required for the uv environment as well, since we'll be running completely unwrapped python when managed by uv.

PR #391, opened 2024-09-05
[ci] make checks faster

ryantm:
Why
===
* GitHub actions is taking way to long to build all our modules.
* Instead, let's just eval them and save the building for CD

What changed
===
* use `nix eval .#bundle` isntead of the build_change_modules script
* Remove some unushed scripts

Test plan
===
* CI passes in less than 15 minutes

Rollout
=======


blast-hardcheese (approved):
:chart_with_downwards_trend:

PR #394, opened 2024-09-17
Exposing global "REPLIT_LD_LOADER" property to be picked up by replenv builder

blast-hardcheese:
Why
===

For better or worse, `uv` mandates the virtualenv being a "real" virtualenv, including `bin/python` symlinks. This subverts our attempts to manipulate the environment via shell-script wrappers, causing runtime library issues.

By moving the environment shuffling into the replenv builder itself, curating the environment itself instead of the wrapper scripts, we can have both uv as well as tools like rtld-loder.

What changed
============

- Expose `REPLIT_LD_AUDIT`

Test plan
=========

I should be able to `echo $REPLIT_LD_AUDIT` and get the value defined even on an empty Repl.

Rollout
=======

### DIFF
--- a/pkgs/modules/replit/default.nix
+++ b/pkgs/modules/replit/default.nix
@@ -34,10 +34,13 @@ in
     Replit tools. Includes .replit language server.
   '';
 
+  replit.env = {
+    REPLIT_LD_AUDIT = "${pkgs.replit-rtld-loader}/rtld_loader.so";
+  };
+
   replit.dev.languageServers.dotreplit-lsp = {
     name = ".replit LSP";
     language = "dotreplit";
     start = "${taplo}/bin/taplo lsp -c ${taplo-config} stdio";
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
I've confirmed that this puts the entry I need in the right place -- the intent is that this will replace not only the python wrappers but also the node/bun wrappers as well.

airportyh:
Question, does this mean we are not setting this var at all for our current python setup?

blast-hardcheese:
@airportyh Currently the value is interpolated into the Python wrapper directly, but I'm trying to move it to "globally available" if the flag is turned on.

Is it OK to be globally enabled for all binaries within a Repl? I would presume yes, at a possible slight impact to startup speed, but effectively a no-op after that, but we'd get the benefit of broken applications and libraries working out of the box without further patching.

PR #396, opened 2024-09-20
Binary python wrapper

blast-hardcheese:
Why
===

`uv` uses some internal tooling (similar to `python -m site`) to determine the python interpreter to use. By using a custom python wrapper, we can get a little more control of that environment and hopefully convince uv to use our wrapper as the canonical python version, bringing in `LD_AUDIT` with a tight scope until https://github.com/jemalloc/jemalloc/issues/2472 has been resolved.

What changed
============

- Add a Go binary, `.python-wrapped`, with relevant symlinks
- Add some miscellaneous quality of life improvements to the Repl environment for this repo

Test plan
=========

`upm -l python3-uv add requests` should create a `.pythonlibs` where `bin/python` points to our `.python-wrapped` binary.

`upm add streamlit`, followed by
`python -c 'import numpy'`
should import cleanly, and
`nix run nixpkgs#python312 -- -c 'import numpy'`
should fail because `LD_AUDIT` was not present.

Rollout
=======

uv is currently blocked, so if there's a need to roll back we should also roll back `flag-agent-uv` if applicable.

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -94,7 +94,7 @@ let
 
   poetry-wrapper = pythonWrapper { bin = "${poetry}/bin/poetry"; name = "poetry"; };
 
-  python-wrapped = pkgs.callPackage ../../python-wrapped {
+  binary-wrapped-python = pkgs.callPackage ../../python-wrapped {
     inherit pkgs python python-ld-library-path;
   };
 
@@ -114,7 +114,7 @@ in
   '';
 
   replit.packages = [
-    python-wrapped
+    binary-wrapped-python
     pip-wrapper
     poetry-wrapper
     pkgs.uv
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -165,7 +165,7 @@ in
     # Even though it is set-default in the wrapper, add it to the
     # environment too, so that when someone wants to override it,
     # they can keep the defaults if they want to.
-    PYTHON_LD_LIBRARY_PATH = python-ld-library-path;
+    REPLIT_PYTHON_LD_LIBRARY_PATH = python-ld-library-path;
     PATH = "${userbase}/bin";
   };
 }
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -94,6 +94,10 @@ let
 
   poetry-wrapper = pythonWrapper { bin = "${poetry}/bin/poetry"; name = "poetry"; };
 
+  python-wrapped = pkgs.callPackage ../../python-wrapped {
+    inherit pkgs python;
+  };
+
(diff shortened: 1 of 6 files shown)


blast-hardcheese:
image
manual testing correctly picks up the interpreter

blast-hardcheese:
image
the wrapped python still correctly picks up our sitecustomize

blast-hardcheese:
importing `numpy` with `.python-wrapped` works, importing `numpy` with stock python fails:
image

pkgs/python-wrapped/default.nix
```diff
@@ -0,0 +1,32 @@
...
+  vendorHash = null;
+
+  meta = with pkgs.lib; {
+    description = ''
+      A replit proxy to Python which interpolates some critical bits
+      of the environment at runtime
+    '';
+    homepage = "https://replit.com";
+    license = licenses.mit;
+    mainProgram = "python-wrapped";
```
ryantm, line 22:
```suggestion
    mainProgram = "python";
```

Pretty sure this should be python, you aren't making a binary available in $out/bin called "python-wrapper".

ryantm (commented):
Looks fine to me. I'd like @airportyh to check the LD audit stuff.

pkgs/modules/python/default.nix
```diff
@@ -94,6 +94,10 @@ let
 
   poetry-wrapper = pythonWrapper { bin = "${poetry}/bin/poetry"; name = "poetry"; };
 
+  python-wrapped = pkgs.callPackage ../../python-wrapped {
```
airportyh, line 97: python-wrapper vs python-wrapped seems too subtly different. Can we make them more different?

airportyh:
Can you test the wrapper with normal poetry setup to make sure it works too? Also of course run templates tests before deploy.

airportyh (approved):

pkgs/python-wrapped/default.nix
blast-hardcheese: Let me give it another shot -- I'm intentionally offering the binary as `.python-wrapped` with three symlinks, as a sign to users digging in that we're engaging in some shenanigans.

pkgs/modules/python/default.nix
blast-hardcheese: Yes, no problem. My intention was to actually delete `python-wrapper` entirely, but it was too deeply entwined with the prybar pythons.

pkgs/python-wrapped/default.nix
ryantm: mainProgram is only used by `nix run` and it just looks in $out/bin for the name to run. If it's not there, it will probably be messed up. That said, it's not likely someone is going to be nix running this so it doesn't really matter.

blast-hardcheese: aah, I thought it was the other way around. OK, I'll make sure it's set.

PR #397, opened 2024-09-20
Moving py3.8 over to historical modules

blast-hardcheese:
Why
===

In preparation for next month's EOL, opening an intent-to-deprecate PR so we can stop building python3.8.

Sphinx [is not supported for python 3.8 and below](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/sphinx/default.nix#L45), and while I cannot find it at this point, there is a long thread of people talking about how it's not worth trying to backport since 3.8 is so close to EOL.

What changed
============

Moving the `python-3.8` module over to historical modules, deleting the module definition.

Test plan
=========

It should be possible to switch from our wrapped pip over to stock pip, which has been blocked by this sphinx issue.

Rollout
=======

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -92,26 +92,6 @@
           };
         };
 
-      } // rec {
-        python38 = pkgs.python310.override {
-          sourceVersion = {
-            major = "3";
-            minor = "8";
-            patch = "18";
-            suffix = "";
-          };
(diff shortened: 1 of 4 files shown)


airportyh (approved):
Thanks! It's time.

ryantm (approved):

PR #405, opened 2024-10-11
bug: Removing debugpy override

blast-hardcheese:
Why
===

Looks like debugpy got bumped upstream to a version whose nixpkgs patches no longer apply to our pinned 1.8.0 version. Looks like https://github.com/replit/nixmodules/pull/87 had pinned due to issues with Python 3.11, but googling around suggests this is no longer an issue.

What changed
============

- Removed debugpy override

Test plan
=========

- Manual testing, adding breakpoints, etc.

Rollout
=======

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -32,21 +32,8 @@ let
     destination = "/config.toml";
   };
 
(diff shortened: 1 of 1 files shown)


ryantm (approved):
I don't remember the reason we had this, but if the debugger still works we're good.

PR #403, opened 2024-10-09
nix flake update

blast-hardcheese:
Why
===

Bump so we can get latest available uv

What changed
============

`nix flake update`'d

Test plan
=========

_Describe what you did to test this change to a level of detail that allows your reviewer to test it_

Rollout
=======

_Describe any procedures or requirements needed to roll this out safely (or check the box below)_

- [ ] This is fully backward and forward compatible

PR #406, opened 2024-10-12
chore/fix the build

blast-hardcheese:
Why
===

Just reverting for now so we can get the pyright-extended upgrade, helping `uv` users. We can try to land this change again next week.

What changed
============

Reverting #405 and #403

Test plan
=========

Does the new nixmodules build?

Rollout
=======

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,3 @@ or
 ```
 nix build .#custom-bundle-squashfs
 ```
-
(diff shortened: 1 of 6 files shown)


blast-hardcheese:
@replit/micromanager merge fix the build please

jackyzha0 (approved):

PR #408, opened 2024-10-16
Avoid incorrectly setting LD_LIBRARY_PATH

blast-hardcheese:
Why
===

When I split out the binary Python wrapper, I misread that `LD_LIBRARY_PATH` should be set in all cases, not just the legacy case.

What changed
============

Increasing readability and addressing the underlying bug. Also removed the `REPLIT_RTLD_LOADER = "1"` check as the flag has been rolled out and burned in, and is unlikely to get rolled back.

Test plan
=========

Attempt to fork and run one of the test cases supplied by support.

Rollout
=======

### DIFF
--- a/pkgs/python-utils/default.nix
+++ b/pkgs/python-utils/default.nix
@@ -32,12 +32,8 @@ let
       ldLibraryPathConvertWrapper = pkgs.writeShellApplication {
         inherit name;
         text = ''
-          if test "''${REPLIT_RTLD_LOADER:-}" = "1" && test "''${REPLIT_NIX_CHANNEL:-}" != "legacy"
+          if test "''${REPLIT_NIX_CHANNEL:-}" = "legacy"
           then
-            # activate RTLD loader!
-            export LD_AUDIT="${pkgs.replit-rtld-loader}/rtld_loader.so"
-            export REPLIT_LD_LIBRARY_PATH=${python-ld-library-path}:''${REPLIT_LD_LIBRARY_PATH:-}
-          else
             export LD_LIBRARY_PATH=${python-ld-library-path}
             if [ -n "''${PYTHON_LD_LIBRARY_PATH-}" ]; then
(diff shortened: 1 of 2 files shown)


pkgs/python-utils/default.nix
```diff
@@ -32,19 +32,19 @@ let
       ldLibraryPathConvertWrapper = pkgs.writeShellApplication {
         inherit name;
         text = ''
-          if test "''${REPLIT_RTLD_LOADER:-}" = "1" && test "''${REPLIT_NIX_CHANNEL:-}" != "legacy"
+          if test "''${REPLIT_NIX_CHANNEL:-}" = "legacy"
```
airportyh, line 35: there might be 2 channels to use, let me check.

airportyh: Try to go by this as well https://github.com/replit/goval/blob/main/pkg/replenv/modules/modules.go#L682

blast-hardcheese:
@airportyh only in `pip` and `poetry` wrappers. Once Python 3.8 is EOL then #397 can be merged and the shell wrapper can be cleaned up.

pkgs/python-utils/default.nix
blast-hardcheese: ah TIL, I guess we have gotten lucky that few users have hit that case?

blast-hardcheese: https://github.com/replit/nixmodules/pull/408/files#diff-d53b1e014cd069c8da29b3a8029151e8f4727a885547eba3ce4422266bca8eecR43

ryantm, line 35: Something like this should resolve it, I think:

```suggestion
          if test "''${REPLIT_NIX_CHANNEL:-}" = "legacy" || test "''${REPLIT_NIX_CHANNEL:-}" = "stable-21_11"
```

ryantm (approved):

PR #417, opened 2024-12-18
Bun update

7heMech:
Why Bun is quickly evolving, we gotta keep ups.
Also would it be possible to push changes to repls once this gets merged since in replit bun is still at .34 and this is .40

### DIFF
--- a/pkgs/bun/default.nix
+++ b/pkgs/bun/default.nix
@@ -3,9 +3,9 @@
 }:
 
 bun.overrideAttrs rec {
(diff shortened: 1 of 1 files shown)


blast-hardcheese (approved):

blast-hardcheese:
@7heMech It's rolling now! Pardon for letting the last PR languish, the process for promoting is currently very manual.

I'll investigate to see if there's something we can do to add more visibility when stuff is just sitting waiting to go out.

PR #419, opened 2024-12-19
[go, nixpkgs, dotnet] chore/bump unstable

blast-hardcheese:
Why
===

We've been pinned to an old version of uv because nixpkgs has been blocked from upgrading by bitrot.

Fortunately now we get the latest `uv`:
```
$ which uv
/nix/store/w7cs06nn9vvk8a59nv48y4biv9zzv43x-uv-0.4.30/bin/uv
```

What changed
============

- Move go_1_21 to historical modules
- Upgrade to `goBuildModule`
- Reflow `dap-python`

Test plan
=========

- Manual testing
- CI
- Staging

Rollout
=======


pkgs/modules/docker/default.nix
```diff
@@ -102,18 +102,18 @@ let
...
     src = pkgs.fetchFromGitHub {
       owner = "moby";
       repo = "moby";
       rev = "v24.0.7";
-      sha256 = "sha256-VUgsclXkoHHNT+GgYL7qiCV/4V3P9RZrT9BegMVYaRU=";
+      hash = "sha256-VUgsclXkoHHNT+GgYL7qiCV/4V3P9RZrT9BegMVYaRU=";
     };
 
-    goPackagePath = mobyGoPackagePath;
+    vendorHash = "";
```
vlinkz, line 116: Needs to be filled in?

blast-hardcheese: I'm kind of confused as to why it wasn't a build failure, either via local `nix build` or CI here. How can this be made to be a build failure, as other required parameters are?

blast-hardcheese: _ah_. I was confused because when the value was absent, it was a statically analyzable error, so it warned me even though I was not building the `docker` module.

When the value was supplied but incorrect, it became a runtime error so it didn't warn me even though it was incorrect.

It didn't flag here because we stopped building the modules as part of PR gating since it was burning so much CPU for little benefit.

blast-hardcheese: And yes, when explicitly built,
```
       > vendor folder exists, please set 'vendorHash = null;' in your expression
```
so all is well.

blast-hardcheese: I did not do this moby upgrade correctly.

blast-hardcheese: #420'd actually.

ryantm (approved):

blast-hardcheese:
Ensured no modules are being removed:

```
11a12
>   "clojure-1.12",
20a22
>   "deno-2",
22a25
>   "dotnet-8.0",
30a34
>   "go-1.23",
47a52
>   "php-8.3",
```

PR #420, opened 2024-12-20
[docker] Punting on docker upgrade

blast-hardcheese:
Why
===

This module requires some rework, so I'm splitting it out from #419 for visibility and revertability.

What changed
============

Comment out docker module

Test plan
=========

Rollout
=======


ryantm (approved):
Yep, that's the trick!

PR #422, opened 2024-12-21
[uv] uv bleeding edge

blast-hardcheese:
Why
===

Bumping uv to latest HEAD in order to get https://github.com/astral-sh/uv/pull/10046, and open us up to just building the latest tags irrespective of what nixpkgs-unstable is tracking.

What changed
============

- Copy uv definition from https://github.com/NixOS/nixpkgs/tree/staging/pkgs/by-name/uv/uv
- Reformulate to use fenix flake input
- Bump uv to get the latest rev

Test plan
=========

Verified that the following builds: ✅
```
$ nix build --no-eval-cache .#'"python-3.12"' -o result.json
```

Verify that transitively depending on `torch` successfully resolves: ✅ 

See the "complete example" in https://docs.astral.sh/uv/guides/integration/pytorch/#using-a-pytorch-index

Rollout
=======

### DIFF
--- a/pkgs/modules/python/uv/default.nix
+++ b/pkgs/modules/python/uv/default.nix
@@ -13,17 +13,17 @@
 
 rustPlatform.buildRustPackage rec {
   pname = "uv";
-  version = "0.5.10";
+  version = "0.5.11"; # Technically 8 versions ahead of 0.5.11
 
   src = fetchFromGitHub {
     owner = "astral-sh";
     repo = "uv";
-    tag = version;
-    hash = "sha256-GE/MgaX6JhzVVwrkz33fr1Vl83QD1WHhlB7vPdJ2W3c=";
+    rev = "ddc290feb4ed2de4740c786af2436cf1f82a3190";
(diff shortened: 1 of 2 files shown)


rmehri01 (approved):
Thanks!

pkgs/modules/python/uv/default.nix
```diff
@@ -0,0 +1,76 @@
...
+, rustPlatform
+, versionCheckHook
+, python3Packages
+, nix-update-script
+,
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "uv";
+  version = "0.5.11"; # Technically 8 versions ahead of 0.5.11
```
ryantm, line 16: What's the story here?

ryantm (approved):
Looks good. Is there a compelling reason to go so bleeding edge?

blast-hardcheese:
@ryantm there's a bugfix merged today that solves an issue with our users' ability to install pytorch. It's still being iterated on by the uv team, but at least now we're in a place to have more control of when new versions drop more broadly.

PR #426, opened 2025-01-17
Update bun

7heMech:
This is probably the last version before 1.2, idk how the migration works on replit then.

### DIFF
--- a/pkgs/bun/default.nix
+++ b/pkgs/bun/default.nix
@@ -3,9 +3,9 @@
 }:
 
 bun.overrideAttrs rec {
-  version = "1.1.40";
+  version = "1.1.44";
   src = fetchurl {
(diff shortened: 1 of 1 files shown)


blast-hardcheese (approved):
lgtm! Thank you!

blast-hardcheese:
To answer your question, when 1.2 gets released, we add an entry with the last `HEAD` from this repo that had the old version into `pkgs/historical-modules/default.nix` (see 71f34ae16def0e7c25bc501bd2ced632239d0d75), then just bump the version to 1.2 as normal.

This persists the last known release of a historical module, but still permits this repository to describe the whole ecosystem of all supported modules, as well as allows us to do hotfixes if some major vulnerability that absolutely must be patched is discovered.

PR #427, opened 2025-02-03
Disable user mode install for pip

airportyh:
Why
===

In some cases, using user install mode for pip will break because with the uv setup we are running in a virtualenv, where user mode will be disabled.

What changed
============

If uv is already rolled out 100% we can consider switching this off. But we may break existing repls that aren't uv. If there's a way to do this only with uv repls, that'll be a better solution.

### DIFF
--- a/pkgs/modules/python/pip.nix
+++ b/pkgs/modules/python/pip.nix
@@ -22,7 +22,7 @@ let
     name = "pip.conf";
     text = ''
       [global]
-      user = yes
+      user = no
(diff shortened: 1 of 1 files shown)


ryantm (approved):
approving, but @blast-hardcheese should probably take a look

blast-hardcheese (approved):
I think this is safe

airportyh:
Oops this breaks pip installs from the shell. I am going to revert.

PR #430, opened 2025-02-04
Fix pip in shell for uv/virtualenv-based python

airportyh:
Why
===

Currently if a repl is based on uv + virtualenv, pip install doesn't work because we have a global `PIP_CONFIG_FILE` setting which forces `--user` option for pip installs, but that fails in virtualenvs. This breaks some packages which use `python -m pip install` internally like https://spacy.io/ but also breaks `python -m pip install` in general.

What changed
============

1. Added a check in `sitecustomize.py` to see if the python executable is coming from the virtualenv `.pythonlibs`. If so: remove the `PIP_CONFIG_FILE` entry from the environment. This allows `python -m pip install ...` to work in both the virturalenv and the non-virtualenv cases.

Test plan
=========

1. Test spacy e2e without virtualenv:
    a. Create a python repl
    b. `rm -fr .pythonlibs`
    c. `pip install -U spacy`
    d. `python -m spacy download en_core_web_sm`
2. Test spacy e2e with virtualenv:
    a. Create a new python repl
    c. `uv sync`
    d. `pip install -U spacy`
    e. `python -m spacy download en_core_web_sm`
3. Test `python -m pip install colorama` with the same above 2 cases.

Rollout
=======

- [ x] This is fully backward and forward compatible

### DIFF
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -162,7 +162,7 @@ in
     POETRY_USE_USER_SITE = "1";
     PIP_CONFIG_FILE = pip.config.outPath;
     PYTHONUSERBASE = userbase;
-    PYTHONPATH = "${sitecustomize}:${pip.sitePackages}";
+    PYTHONPATH = "${sitecustomize}:${pip.pip}/${python.sitePackages}";
     REPLIT_PYTHONPATH = "${userbase}/${python.sitePackages}:${pypkgs.setuptools}/${python.sitePackages}";
     UV_PROJECT_ENVIRONMENT = "$REPL_HOME/.pythonlibs";
     UV_PYTHON_DOWNLOADS = "never";
--- a/pkgs/modules/python/default.nix
+++ b/pkgs/modules/python/default.nix
@@ -162,7 +162,7 @@ in
     POETRY_USE_USER_SITE = "1";
     PIP_CONFIG_FILE = pip.config.outPath;
     PYTHONUSERBASE = userbase;
-    PYTHONPATH = "${sitecustomize}:${pip.pip}/${python.sitePackages}";
+    PYTHONPATH = "${sitecustomize}:${pip.sitePackages}";
     REPLIT_PYTHONPATH = "${userbase}/${python.sitePackages}:${pypkgs.setuptools}/${python.sitePackages}";
     UV_PROJECT_ENVIRONMENT = "$REPL_HOME/.pythonlibs";
     UV_PYTHON_DOWNLOADS = "never";
--- a/pkgs/modules/python/pip.nix
+++ b/pkgs/modules/python/pip.nix
@@ -18,22 +18,6 @@ let
     ];
   });
 
-  # This pip wrapper will detect if python is called from within
-  # the virtualenv: `.pythonlibs`. If so, it will simply call `python -m pip ...`
-  pip-wrapper = pkgs.writeShellApplication {
-    name = "pip";
-    text = ''
-      python_location=$(which python)
-
-      if [ "$python_location" = "''${REPL_HOME}/.pythonlibs/bin/python" ]; then
-        exec python -m pip "$@"
-      else
-        unset PYTHONNOUSERSITE
-        exec "${pip}/bin/pip"  "$@" 
-      fi
-    '';
-  };
-
   config = pkgs.writeTextFile {
     name = "pip.conf";
     text = ''
@@ -45,7 +29,5 @@ let
   };
 in
 {
-  pip = pip-wrapper;
-  sitePackages = "${pip}/${python.sitePackages}";
-  inherit config;
+  inherit pip config;
 }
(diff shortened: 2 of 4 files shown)


pkgs/modules/python/default.nix
```diff
@@ -162,7 +162,7 @@ in
     POETRY_USE_USER_SITE = "1";
     PIP_CONFIG_FILE = pip.config.outPath;
     PYTHONUSERBASE = userbase;
-    PYTHONPATH = "${sitecustomize}:${pip.pip}/${python.sitePackages}";
+    PYTHONPATH = "${sitecustomize}:${pip.sitePackages}";
```
blast-hardcheese, line 165: Why not

```suggestion
    PYTHONPATH = "${sitecustomize}:${python.sitePackages}";
```

?

pkgs/modules/python/pip.nix
```diff
@@ -18,6 +18,41 @@ let
...
   });
 
+  # This pip wrapper will detect if python is called from within
+  # the virtualenv: `.pythonlibs/bin`. If so, it will install a shell script
+  # called `pip` into the virtualenv so that if pip is called in the future
+  # python will think it's in the virtualenv.
+  pip-wrapper = pkgs.writeShellApplication {
+    name = "pip";
+    text = ''
+      python_location=$(which python)
```
blast-hardcheese, line 28:
```suggestion
      python_location="$(which python)"
```

pkgs/modules/python/default.nix
airportyh: Pip's lib code is separated into a separate directory in the nix setup.

pkgs/modules/python/pip.nix
```diff
@@ -18,6 +18,41 @@ let
...
+      venv_pip_location="''${REPL_HOME}/.pythonlibs/bin/pip"
+      venv_pip3_location="''${REPL_HOME}/.pythonlibs/bin/pip3"
+
+      function init_venv_pip {
+        cat <<EOF > "''${venv_pip_location}"
+      #! /bin/sh
+      export PIP_CONFIG_FILE=
```
blast-hardcheese, line 35: I tried this, I think we still need most of the properties in `pip.conf`, just not `user = false`

pkgs/modules/python/pip.nix
```diff
@@ -18,6 +18,41 @@ let
...
+
+      if [ "$python_location" = "''${REPL_HOME}/.pythonlibs/bin/python" ]; then
+        if [ ! -f "''${venv_pip_location}" ]; then
+          init_venv_pip
+        fi
+        exec "''${venv_pip_location}" "$@"
+      else
+        unset PYTHONNOUSERSITE
+        exec "${pip}/bin/pip"  "$@" 
+      fi
```
blast-hardcheese, line 52: Do we want to preserve this legacy path?

blast-hardcheese (approved):
hard to maintain the matrix of different environments we support. This looks OK. We should definitely template-tester as well as verify agent uv flow

pkgs/modules/python/default.nix
blast-hardcheese: right, but do we not need the `python.sitePackages` to have access to (eg) setuptools?

blast-hardcheese: oh wow I guess we really do not have entries for that on `PYTHONPATH` today:

```
~/workspace$ tr : '\n' <<< $PYTHONPATH 
/nix/store/wjh5dp8cqvlaxs7yxmy9amn19d54r9nb-sitecustomize/lib/python/site-packages
/nix/store/dfg63lldbvcj207iyr9z0xw21d8ax02n-python3.11-pip-24.0/lib/python3.11/site-packages
```

airportyh: I see the confusion. `python.sitePackages` actually only yields the string: `lib/python3.11/site-packages`

pkgs/modules/python/pip.nix
airportyh: These?
```
disable-pip-version-check = yes
break-system-packages = yes
```
Why?

pkgs/modules/python/pip.nix
airportyh: Yes. It is used when we are not using uv/virtualenv and still using poetry.

pkgs/modules/python/pip.nix
blast-hardcheese: `break-system-packages = yes` is necessary if pip is running outside of the venv, but I guess it doesn't apply in this case anymore

blast-hardcheese: Resolving, feel free to proceed

pkgs/modules/python/pip.nix
blast-hardcheese: ack

airportyh:
@blast-hardcheese After finding an edge case that didn't work with the logic in `main.go` because the resulting binary getting symlinked by uv into `.pythonlibs/bin`, I ended up with a different approach based on just a few lines in `sitecustomize.py`. Please re-review.

blast-hardcheese (approved):
This seems good

PR #439, opened 2025-03-11
Bun update

7heMech:
The Bun keeps getting hotter, and we should not turn the oven off.


ryantm (approved):
> The Bun keeps getting hotter, and we should not turn the oven off.

:fire:

PR #442, opened 2025-04-26
Bump pyright-extended to 2.0.13

daweifeng-replit:
Why
===

The latest pyright-extened version is 2.0.13 and it should ignore non-python files so we should see less `ENOSPC: System limit for number of file watchers reached` errors. 

pyright-extended PR: https://github.com/replit/pyright-extended/pull/66

What changed
============

- Bump version to 2.0.13

Test plan
=========

Rollout
=======


ryantm (approved):
This change looks good but let's test a lot since the LSP is pretty critical!

PR #454, opened 2025-07-18
fix bundle build

vlinkz:
Why
===

Add some fixes that are causing the bundle to fail to build.

What changed
============

- Remove go 1.23, make auto update to go 1.24
- Update hashes
- Remove python3.14 (was added in last pr), as pip is broken in nixpkgs for now

Test plan
=========

Build bundle

Rollout
=======


ryantm (approved):
> As always, the release maintains the Go 1 [promise of compatibility](https://tip.golang.org/doc/go1compat). We expect almost all Go programs to continue to compile and run as before.

PR #455, opened 2025-07-18
bump rtld.optional_static_tls

vlinkz:
Why
===

Latest rustc nightly fails with rtld loader enabled with the current static tls limit.

What changed
============

Bump rtld.optional_static_tls from 2500 -> 10000.

Test plan
=========

`rustc` nightly and other programs still work as expected.

Rollout
=======


ryantm (approved):
What's another 7.5kb among friends?

